PR #20, opened 2023-03-08
chore: set up mongodb connection

### LINKED ISSUE #19: set up mongo database

arrocke:
closes #19


Pertempto:
What is your branch naming scheme? I noticed the `build/` at the beginning.

Pertempto:
I'm guessing I need to do some things to set up a mongo DB in my local environment?  
image

arrocke:
Yeah, you are going to need to install mongodb v5.

For branch names, at my job we usually do `<type>/<story #>-<description>`, but for this I don't think we need a naming convention. Branch names don't stick around the history, but commits do, so I care more about them

Pertempto:
> Yeah, you are going to need to install mongodb v5.

That information should be added to `README.md` in the Setup section, for future developers.   

I ran the installer with the default options, except I didn't install the MongoDB router because it took up extra storage space and I didn't think I needed it.  
We really should have more detailed install steps than "Install mongodb v5" in the `README`

Pertempto:
Just curious, why are we using MongoDB v5 when v6 is the latest version?  
It seems strange to start a fresh project with out-of-date dependencies.

packages/api/pages/api/health.ts
```diff
@@ -1,5 +1,10 @@
 import { NextApiRequest, NextApiResponse } from 'next'
+import { client } from '../../db'
 
-export default function(req: NextApiRequest, res: NextApiResponse) {
-  return res.status(200).send('ok')
+export default async function(req: NextApiRequest, res: NextApiResponse) {
+  const collections = await client.db().listCollections().toArray()
+  return res.status(200).json({
+    status: 'ok',
+    collections
```
Pertempto: I assume we're just doing this temporarily for debugging in the MVP stage? It doesn't seem ideal to broadcast data straight from the DB on a non-authenticated API route.

Pertempto (changes requested):
I was able to get MongoDB installed and the code running.  
I think good setup documentation is very essential, otherwise it takes a lot of developer time to on-board new contributors. We want to document all the details required to install dependencies and setup a developer's local environment.

packages/api/pages/api/health.ts
arrocke: Yep, I just wanted something to demonstrate that the database connection was working.

arrocke:
> Just curious, why are we using MongoDB v5 when v6 is the latest version?

MongoDB v6 was released last august, and is not available for free hosting on Atlas. Eventually we will have to switch to paid db hosting, at which point we can upgrade. The [v5 - v6 upgrade](https://www.mongodb.com/docs/manual/release-notes/6.0-compatibility/) is not likely to be super involved for our use case. Other than the removal of already deprecated features, it mostly impacts the hosting provider.

> We want to document all the details required to install dependencies and setup a developer's local environment.

Great point. I will add this before I merge the PR.

> It doesn't seem ideal to broadcast data straight from the DB on a non-authenticated API route.

I'll change this to return a boolean based on whether that command can complete or not. That way, we can determine whether the db connection is healthy or not without exposing anything to unauthenticated users.

Pertempto:
I deleted the `gloss-translation` database in my Mongo Compass, but the health endpoint still says the connection is ok. Is that expected? I was trying to test what it would do if the DB connection failed.

arrocke:
Yep, that's expected. In mongo you connect to the server, which can have multiple databases in it. When you execute a query against a database that doesn't exist it pretends it does, and when you insert a document into that database it creates the database and collection on the fly. To properly test, you'd want to either stop the db server, or change the connection url to something that isn't running mongodb

Pertempto (approved):
It looks good to me.

PR #34, opened 2023-03-15
chore: app layout

### LINKED ISSUE #21: organize web app
## acceptance criteria
- set up router
- set up directory structure for the following features
  - shared code like base components, api helpers, etc
  - user management
  - language management
  - translation interface

arrocke:
Here is what I'm suggesting we use for the file structure for the web project. I find grouping things by domain to be easier to manage and find than by file type. It also helps in component design because shared files need to be more generic, while domain specific files can have more specialization.
```
src/
  app/ - top level files (router, layout components, etc)
  shared/ - components and hooks that are shared across features
  features/ - views and components grouped by major feature
    translation/
    languages/
    users/
```

We will definitely want to document this for new contributors, but I think I'd like to wait a bit until the dust settles on the MVP so that we aren't doing unnecessary work while we figure things out

closes #21


vercel.json
arrocke, line 1: I was able to figure out how to proxy api requests. There are three rewrites here that vercel will check on every request.
1. First check if we are on an api host. `gloss-translation-api.*` will match any api server in both production and preview. We need this because both servers share the same config file, so this rewrite ensures that none of the others match.
2. Then check if we are making an api request from a web server. `gloss-translation(?<host>-.+|$)` will match both production and preview web servers and grab the random suffix for preview servers. Then we can use that suffix to build the correct api hostname.
3. Finally, check for all other routes that don't have a `.` in them and rewrite them to the index file. This supports the client side routing in react

packages/web/src/app/NotFound.tsx
Pertempto, line 1: I always like friendly 404 pages 👍🏽

Pertempto (approved):
Everything runs well in my local environment, and I don't see any concerns in the code.

PR #36, opened 2023-03-20
build: set up prisma

### LINKED ISSUE #35: second guessing mongodb
I had a conversation with a coworker about NoSQL vs SQL today that has me second guessing our decision to go with mongo. If we were to use a SQL database, it would probably look something similar to this:

```mermaid
erDiagram
  BOOK ||--|{ CHAPTER : contains
  CHAPTER ||--|{ VERSE : contains
  VERSE ||--|{ WORD : contains
  WORD }|--|| FORM: has
  FORM }|--|| LEMMA : has
  LANGUAGE ||--|{ GLOSS : contains
  GLOSS ||--|| WORD : references

  BOOK {
    number id PK
    string name
  }

  CHAPTER {
    number id PK
    number bookId FK
    number number 
  }

  VERSE {
    number id PK
    number chapterId FK
    number number
  }

  WORD {
    number id PK
    number verseId FK
    ID lemmaId FK
    string text
  }

  FORM {
    ID id PK
    ID lemmaId FK
    string text
    string grammarCode
  }

  LEMMA {
    ID id PK
    string text
    number strongs
  }

  GLOSS {
    ID id PK
    ID languageId FK
    ID wordId FK
    string gloss
  }

  LANGUAGE {
    ID id PK
    string name
  }
```

If that is the case, then to query all glosses for a lemma would be a query something like this. My SQL is a bit rusty so this might not be exactly right.
```sql
SELECT Book.name, Chapter.number, Verse.number, Word.id, Gloss.text, Form.text
FROM Form
JOIN Word On Word.formId=Form.id
JOIN Gloss On Gloss.wordId=Word.id
JOIN Verse On Word.verseId=Verse.id
JOIN Chapter On Verse.chapterId=Chapter.id
JOIN Book On Chapter.bookId=Book.id
WHERE Form.lemmaId=<lemma id> AND Gloss.languageId=<language id>
```

Now that is a lot of joins, but keep in mind that there are 443,000 words in the ancient languages. And besides Yaweh and God, all other words are used less than 3000 times. So that means the lookup of words is going to pull out in most cases <3000 words from 443,000 and then join that against the verse/chapter/book tables which will all be small, and against the gloss table which if we index by language id should be fast as well regardless of how many languages we end up supporting. I think if we were to look at fetching a verse with its glosses and grammar information, it would be similar.

Now, for mongodb to give us the benefit of speed we have to store the gloss information twice, once with the verse and once with the lemma and form. You can't rely on joins for speed in mongo.

Now the key question that I have is whether it is worth the added complexity to denormalize our data with mongo in order to get some perceived speed boost in queries? The reason I went through that exercise is because I'm really wondering whether our data is going to be large enough even with 40+ languages for a sql server to struggle to deliver this data in a timely manner. I'm not super familiar with SQL, so maybe this is rather obvious. But if sql can handle our data, then we should not saddle ourselves with the added complexity and constraints of a nosql database.

arrocke:
closes #35


packages/api/.env
```diff
@@ -1 +1 @@
-MONGODB_URI=mongodb://127.0.0.1:27017/gloss-translation
\ No newline at end of file
+DATABASE_URL=
```
arrocke, line 1: You'll have to add a `.env.local` file to set up your postgres database

packages/api/prisma/schema.prisma
```diff
@@ -0,0 +1,38 @@
...
+}
+
+model Verse {
+  id        Int     @id
+  number    Int
+  chapter   Chapter @relation(fields: [chapterId], references: [id])
+  chapterId Int
+
+  @@unique([chapterId, number])
+}
```
arrocke: Here is what the data model looks like for prisma. This is used to generate the database migrations every time it is modified, and to generate the typescript client

packages/api/prisma/migrations/20230319033722_/migration.sql
```diff
@@ -0,0 +1,41 @@
+-- CreateTable
+CREATE TABLE "Book" (
+    "id" INTEGER NOT NULL,
+    "name" TEXT NOT NULL,
+    "search" TEXT NOT NULL,
+
+    CONSTRAINT "Book_pkey" PRIMARY KEY ("id")
+);
```
arrocke, line 8: This migration was generated by prisma and is ready to be deployed to a production database

packages/api/project.json
```diff
@@ -50,6 +50,36 @@
       "options": {
         "lintFilePatterns": ["packages/api/**/*.{ts,tsx,js,jsx}"]
       }
+    },
+    "prisma-generate": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "npx prisma generate --schema=packages/api/prisma/schema.prisma"
+      }
+    },
```
arrocke, line 59: Two major things have to be adjusted from the standard set up so that the prisma schema and client can be within the package rather than in the root of the repo. First we need to define the common prisma commands in the project file. This solves a few problems.

1. Prisma expects the .env file to be stored in the root of the repo. If nx executes prisma, then it can load the environment for us from the .env file in the api package
2. Prisma expects the schema to be at prisma/schema.prisma. These commands allow us to set that path so we don't have to type that in every time we want to execute a prisma command.

The downside is that for every prisma command you want to use, you have to add it here rather than having one that you can pass multiple arguments to. This is because nx and prisma handle argument parsing differently.

packages/api/prisma/schema.prisma
```diff
@@ -0,0 +1,38 @@
+// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+generator client {
+  provider = "prisma-client-js"
+  output   = "./client"
+}
```
arrocke, line 7: Second, we need to output the client to a directory within the api package rather than in the node modules directory which is the default. This allows multiple packages to have their own prisma client tailored to their own database. Instead of importing `@prisma/client` we just have to import from `./prisma/client` within the api package

packages/api/.gitignore
```diff
@@ -0,0 +1 @@
+prisma/client
```
arrocke, line 1: Because the prisma client is generated code, we are preventing it from being committed to version control

packages/api/pages/api/verse.ts
```diff
@@ -0,0 +1,39 @@
...
+          number: true,
+          book: {
+            select: {
+              name: true
+            }
+          }
+        }
+      }
+    }
+  })
```
arrocke, line 28: And here is it in use. We can select a verse based on its foreign keys in multiple tables, and prisma can decide how it wants to build the query or queries it needs. Prisma exposes higher level query functions rather than trying to map exactly to SQL. This query here actually results in three queries for the verse, chapter, and book. We can always drop into a SQL command if we want to though.

package.json
```diff
@@ -2,24 +2,28 @@
   "name": "gloss-translation",
   "version": "0.0.0",
   "license": "MIT",
-  "scripts": {},
+  "scripts": {
+    "postinstall": "nx run-many --target=prisma-generate --all"
```
arrocke, line 6: This script ensures that any packages that need a prisma client have one generated

arrocke (commented):
This obviously won't work on vercel yet until we've got a postgres database, but here you can test it locally. Once we've got a database on vercel, we can set up automated database migrations by changing the vercel build step to `npx nx prisma-deploy api && npx nx build api --prod`

For now, to test locally:
1. Set up a postgresql14 database and add the connection string to your .env.local file in the api package
2. Run `npm i` to install the prisma packages and generate the prisma client
3. Run `npx nx prisma-migrate api` to update your local database to match the prisma schema
4. In a SQL terminal, run the following commands to get some test data scaffolded:
```sql
INSERT INTO public."Book" (id, name, search) VALUES (40, 'Matthew', 'matthew');
INSERT INTO public."Chapter" (id, number, "bookId") VALUES (1, 1, 40);
INSERT INTO public."Verse" (id, number, "chapterId") VALUES (1, 1, 1);
```
5. Start the server and go to `http://localhost:4300/api/verse` to see prisma in action

The next thing for me to figure out is how to seed the db with the basic verse data necessary for starting a development server.

arrocke:
I'm really happy with this, with the exception of the number of nx commands we have to set up. We might be able to create an nx plugin to make that easier if/when we have multiple databases, but for now this should work.

I really like the prisma is responsible for the schema and all database and typescript changes stem from that. The client seems pretty good although I find it to be a bit more restrictive than the mongo client, but the type safety is much better.

packages/api/prisma/schema.prisma
```diff
@@ -0,0 +1,38 @@
...
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+model Book {
+  id      Int       @id
+  name    String
+  search  String    @unique
```
Pertempto: What is this field for?

arrocke: I'm guessing we are going to want to differentiate between the display text for a book which will be in multiple languages from the string used to look up the book. This will make it easy for us to convert `matthew 1:1` into a database query. We can then expect the frontend to generate a reference that follows that pattern regardless of the display language, rather than having to know that id of the book. Ultimately, I think we are going to want API routes that look something like one of these:
```
GET /book/matthew/chapter/1/verse/1
GET /verses?reference=matthew1:1
```

arrocke:
@Pertempto With what you've seen here, does this address your concerns with using a SQL database? Or are there issues you'd like to see resolved before we proceed?

packages/api/prisma/schema.prisma
Pertempto: Ok, that sounds great. It's not immediately clear from reading the schema file though, so it would be good to include an explanation comment for that field.

Pertempto:
> @Pertempto With what you've seen here, does this address your concerns with using a SQL database? Or are there issues you'd like to see resolved before we proceed?

Yeah, the code looks fine and I think Prisma will fulfill our requirements well. I'm still working on installing postgres, so I haven't been able to test the code yet.

packages/api/.env
Pertempto: Am I supposed to leave `.env` with these contents? I created a `.env.local` file with this content:  
```
DATABASE_URL=localhost:5432
```
But it seems to load configuration from both files, which leads to an error in compiling  
```
info  - Loaded env from C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\packages\api\.env.local
info  - Loaded env from C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\packages\api\.env
event - compiled client and server successfully in 25.9s (150 modules)
[ ready ] on http://localhost:4300
wait  - compiling /_error (client and server)...
event - compiled client and server successfully in 885 ms (151 modules)
```

arrocke: That should be `DATABASE_URL=postgresql://localhost:5432/db-name`.

You can leave .env alone since .env.local takes precedence. So we can put in .env vars that aren't secret, and .env.local ones that shouldn't be checked in

PR #39, opened 2023-03-28
docs: api routes for translation languages

### LINKED ISSUE #18: api/db design for languages
## acceptance criteria
* sketch out API routes for the following routes
  * GET languages - get all languages
  * GET language - get a single language by its language code
  * POST language - create a new language with a name and code
  * PATCH language - change the language name or code
* sketch out db model to support these routes

arrocke:
closes #18


packages/api/pages/explorer/index.tsx
```diff
@@ -0,0 +1,8 @@
+import dynamic from 'next/dynamic';
+import 'swagger-ui-react/swagger-ui.css';
+
+const SwaggerUI = dynamic(() => import('swagger-ui-react'), { ssr: false });
+
+export default function ApiDocsView() {
+  return <SwaggerUI url="/spec/index.yml" />;
```
arrocke, line 7: This UI belongs on the server side because it is docs for that API. That also prevents normal users from needing to download these resources when they will never see them in the app.

packages/api/public/spec/index.yml
```diff
@@ -0,0 +1,21 @@
...
+    description: Local server
+paths:
+  /languages:
+    $ref: ./paths/languages.yml
+  /languages/{id}:
+    $ref: ./paths/language.yml
+components:
+  schemas:
+    Language:
+      $ref: ./schemas/language.yml
```
arrocke, line 21: This is the root open api spec, and we can import everything else from other files. Unfortunately, there is no good ways for having intermediate references to do things like collect all of the language paths in a single reference, so instead every component and path has to be referenced separately.

packages/api/public/spec/schemas/language.yml
```diff
@@ -0,0 +1,23 @@
+type: object
+properties:
+  type:
+    type: string
+    enum: [language]
+  id:
+    type: string
+    example: es
+    description: The [IETF language code](https://en.wikipedia.org/wiki/IETF_language_tag).
```
arrocke, line 9: I'm thinking it would be good to follow something like [JSON:API](https://jsonapi.org/) for our API design since that will make things consistent, discoverable, and take a lot of the decision making out of the design process.

That's what I've mocked up here. We can definitely revist this if you think this format adds too much unnecessary complexity.

packages/api/prisma/schema.prisma
```diff
@@ -36,3 +38,9 @@ model Verse {
 
   @@unique([chapterId, number])
 }
+
+model Language {
+  id   String @id @default(dbgenerated("generate_ulid()")) @db.Uuid
+  code String @unique
```
arrocke, line 44: Two comments here:
1. I'm a big fan of UUIDs for primary keys for a variety of reasons, but they have a problems with indexing. [A ULID](https://blog.daveallie.com/ulid-primary-keys) is more deterministic because it has a timestamp which solves some of these problems.
2. Even though it is unlikely for a code to be changed, I'm still inclined to have an opaque ID to use for foreign keys.

packages/api/prisma/migrations/20230328215901_add_languages/migration.sql
```diff
@@ -0,0 +1,19 @@
+-- CreateExtension
+CREATE EXTENSION IF NOT EXISTS "pgcrypto";
+
+CREATE OR REPLACE FUNCTION generate_ulid() RETURNS uuid
+    AS $$
+        SELECT (lpad(to_hex(floor(extract(epoch FROM clock_timestamp()) * 1000)::bigint), 12, '0') || encode(gen_random_bytes(10), 'hex'))::uuid;
+    $$ LANGUAGE SQL;
```
arrocke, line 7: I swiped this from that [same page](https://blog.daveallie.com/ulid-primary-keys)

packages/api/prisma/schema.prisma
```diff
@@ -2,13 +2,15 @@
...
+  output          = "./client"
+  previewFeatures = ["postgresqlExtensions"]
 }
 
 datasource db {
-  provider = "postgresql"
-  url      = env("DATABASE_URL")
+  provider   = "postgresql"
+  url        = env("DATABASE_URL")
+  extensions = [pgcrypto]
```
arrocke, line 13: This extension is necessary for that ulid function

packages/api/prisma/schema.prisma
Pertempto: Interesting, I had never heard of ULIDs before. Thanks for the article link!

PR #40, opened 2023-03-30
feat: basic language management

### LINKED ISSUE #31: interface to add languages
## acceptance criteria
- link to a page from the header
- show list of current languages
- button to add language
- show modal to set the language name

arrocke:
## Left To Do
- [x] validate bodies of API requests
- [x] handle HTTP errors on the frontend
- [x] validate forms before submission
- [x] configure API proxy for nested routes

closes #31


package.json
```diff
@@ -56,6 +56,7 @@
     "react-refresh": "0.10.0",
     "tailwindcss": "3.2.7",
     "typescript": "4.9.5",
-    "url-loader": "4.1.1"
+    "url-loader": "4.1.1",
+    "zod": "3.21.4"
```
arrocke, line 60: zod is great for validation with typescript. We can use this to ensure that the HTTP request body matches the typescript type before continuing with the request logic

packages/api-types/src/language.ts
```diff
@@ -0,0 +1,43 @@
...
+  data: LanguageWithLinks;
+}
+
+export interface PostLanguageRequestBody {
+  data: Language;
+}
+
+export interface PostLanguageResponseBody {
+  data: LanguageWithLinks;
+}
```
arrocke, line 35: The api-types package will contain all of the types for all of the HTTP routes. The api project can use these types to ensure that the routes conform to these types, and the api-client package can use these types to cast the JSON received by HTTP requests.

packages/api-client/src/client.ts
```diff
@@ -0,0 +1,97 @@
...
+    Object.setPrototypeOf(this, ApiClientError.prototype);
+  }
+}
+
+export default class ApiClient {
+  readonly languages: Languages;
+
+  constructor(private options: ApiClientOptions = { baseUrl: '' }) {
+    this.languages = new Languages(this);
+  }
```
arrocke, line 55: This client can be used like this:
```js
const languages = await client.languages.findMany()
```

packages/api-client/src/client.ts
```diff
@@ -0,0 +1,97 @@
...
+    return await this.request({ ...request, method: 'GET' });
+  }
+
+  async post<Response>(request: ApiClientPostOptions): Promise<Response> {
+    return await this.request({ ...request, method: 'POST' });
+  }
+
+  async patch<Response>(request: ApiClientPatchOptions): Promise<Response> {
+    return await this.request({ ...request, method: 'PATCH' });
+  }
```
arrocke, line 96: The basic idea is that the client has these helper methods which each part of the client can use to define methods for all of the HTTP routes.

packages/api-client/src/client.ts
```diff
@@ -0,0 +1,97 @@
...
+export type ApiClientPatchOptions = Pick<
+  ApiClientRequestOptions,
+  'path' | 'query' | 'body'
+>;
+export type ApiClientPostOptions = Pick<
+  ApiClientRequestOptions,
+  'path' | 'query' | 'body'
+>;
+
+export class ApiClientError extends Error {
```
arrocke, line 31: Then this error will make it easy to capture errors in frontend code, and the error itself will be typed with an error code and any future fields on errors

packages/api/db.ts
```diff
@@ -1,3 +1,5 @@
-import { PrismaClient } from './prisma/client'
+import { PrismaClient, Prisma } from './prisma/client';
 
-export const client = new PrismaClient()
\ No newline at end of file
+export { Prisma };
```
arrocke, line 3: If we export the Prisma namespace, with the client then we can get access to all of the generated Prisma types for all of the tables

packages/api/helpers.ts
```diff
@@ -0,0 +1,8 @@
+import { ErrorResponse } from '@translation/api-types';
+import { NextApiRequest, NextApiResponse } from 'next';
+
+export type ApiRequest<Params = void> = Params extends void
+  ? NextApiRequest
+  : NextApiRequest & { query: Params };
+
+export type ApiResponse<Body = void> = NextApiResponse<Body | ErrorResponse>;
```
arrocke, line 8: These helpers will make it easier to type the router parameters, and to allow the response json to include error responses along side with the 200 response.

packages/api/pages/api/languages/index.ts
```diff
@@ -1,7 +1,101 @@
...
+  data: languageSchema(),
+});
+
+export default async function (
+  req: ApiRequest,
+  res: ApiResponse<GetLanguagesResponseBody | PostLanguageResponseBody>
+) {
+  switch (req.method) {
+    case 'GET': {
+      const languages = await client.language.findMany();
```
arrocke, line 22: There is a lot in these routes that could be DRY'd up across routes, but I'm going to save that for another PR. The basic flow here is:

1. Routes in nextjs are tied to the file path
2. In the handler, we first branch based on the method, and return 405 for unsupported methods
3. If the request has a body, we use zod to parse and validate it, returning 409 or 422 depending on the error according to JSON:API
4. Then we use Prisma to either fetch or create resources
5. Then return the response with the right code depending on if any errors were thrown.

packages/api/public/spec/paths/language.yml
```diff
@@ -12,7 +12,7 @@ get:
     "200":
       description: Ok
       content:
-        application/vnd.api+json:
+        application/json:
```
arrocke, line 15: I had to revert to `application/json` which isn't what JSON:API expects, but nextjs didn't have a good way to handle a custom content-type

packages/web/src/shared/components/Link.tsx
```diff
@@ -0,0 +1,10 @@
+import { Link as BaseLink, LinkProps } from 'react-router-dom';
+
+export function Link({ className, ...props }: LinkProps) {
+  return (
+    <BaseLink
```
arrocke, line 5: All of these generic shared components should help us keep a consistent style as we build new views

packages/web/src/public/locales/en/translation.json
```diff
@@ -1,4 +1,12 @@
 {
-  "languagesView": "this is where the languages view will go",
-  "translationView": "translation interface is work in progress"
+  "translationView": "translation interface is work in progress",
+  "language_one": "language",
+  "language_other": "languages",
```
arrocke, line 4: i18next is really well designed. You can define a translation key (ie: 'language'), with modifiers (ie '_one' or '_other' for plural modifiers). Then when you use the key you can pass in a count (id: `t('languages', { count: 10 })`), and it will pick the right modifier. Different languages can have different modifiers. So one language might define modifiers for 'zero', 'one', 'two', 'few' and 'many', while another might just have 'one', and 'many'.

packages/web/src/features/languages/LanguageView.tsx
```diff
@@ -1,11 +1,60 @@
...
+
   const { t } = useTranslation();
 
   return (
-    <div className="absolute w-full h-full flex items-center justify-center">
-      {t('languagesView')}
-    </div>
+    <View fitToScreen>
+      <div className="m-auto w-fit">
+        <ViewTitle>{capitalize(t('language', { count: 100 }))}</ViewTitle>
```
arrocke, line 32: In the view title, we want the plural for many languages, so we pass a large count.

packages/web/src/features/languages/LanguageView.tsx
```diff
@@ -1,11 +1,60 @@
...
+        <List>
+          <ListHeader>
+            <ListHeaderCell className="min-w-[240px]">
+              {t('language', { count: 1 }).toUpperCase()}
```
arrocke, line 36: But in the table header, we want the singular, so we pass count of 1.

vercel.json
```diff
@@ -1,15 +1,5 @@
...
-    {
-      "source": "api/:path*",
-      "destination": "api/:path",
-      "has": [
-        {
-          "type": "host",
-          "value": "gloss-translation-api.*"
-        }
-      ]
-    },
```
arrocke, line 12: This rewrite doesn't turn out to work for nested paths on the api, so I had to refactor these rewrites again

PR #49, opened 2023-04-21
feat: database seed script

### LINKED ISSUE #44: database seed for verse data

arrocke:
closes #44


packages/api/prisma/schema.prisma
```diff
@@ -15,13 +15,12 @@ datasource db {
 
 model Book {
   id     Int     @id
-  name   String
-  search String  @unique // This gives us a human readable key to look up books from URL params.
```
arrocke, line 19: Since the book names are going to be provided by the internationalization strings, we only need a string that we use for API calls and database lookups. I replaced `search` with `name` since that is a little clearer.

packages/api/prisma/schema.prisma
```diff
@@ -15,13 +15,12 @@ datasource db {
...
+  name   String  @unique
   verses Verse[]
 }
 
 model Verse {
-  id           Int                @id
+  id           String             @id
```
arrocke, line 30: I decided to make the verse ID follow a pattern which makes String a better fit. That pattern is: `BBCCCVVV` (B - book id, C - chapter number, V - verse number). Words use the same pattern but add two digits for the word order, which means we don't need an order column to sort words since the ID already has that information.

arrocke (commented):
`nx seed api` can be used to load data into the database. It takes about 2 minutes on my computer. So to set up a new database, you run `nx prisma-reset api` and then `nx seed api`.

packages/api/prisma/schema.prisma
Pertempto: I agree, that is much clearer.

packages/api/prisma/schema.prisma
Pertempto: Alright! Where are we documenting this ID pattern?

docs/db.md
Pertempto, line 1: Thanks for adding this.

packages/api/prisma/seed.ts
```diff
@@ -0,0 +1,129 @@
...
+            id: wordId,
+            text,
+            verseId: verse.id,
+            grammar,
+            strongs,
+          });
+
+          if (!lemmas[strongs]) {
+            lemmas[strongs] = {};
+          }
```
Pertempto:
```suggestion
          lemmas[strongs] ??= {};
```
Would this work?

packages/api/prisma/seed.ts
```diff
@@ -0,0 +1,129 @@
...
+          if (!lemmas[strongs][grammar]) {
+            lemmas[strongs][grammar] = { verseIds: [] };
+          }
```
Pertempto:
```suggestion
          lemmas[strongs][grammar] ??= { verseIds: [] };
```

Pertempto (commented):
It all looks pretty good to me. I don't have time to test functionality right now, but I should be able to this evening.

packages/api/prisma/schema.prisma
arrocke: I think we should start a docs file for the database structure. I do document this in the open api specs in the other PR, but I think it would be helpful to document the database schema itself. Maybe there is a way to generate this from the prisma file as well

packages/api/prisma/seed.ts
arrocke: I don't see why not! I like it

packages/api/prisma/schema.prisma
```diff
@@ -10,62 +10,102 @@ generator client {
...
 }
 
+/// A book of the Bible.
 model Book {
+  /// The ID of the book. Corresponds to its order in the Bible.
   id     Int     @id
+  /// The three letter code used to for look ups from references.
   name   String  @unique
+  /// The list of verses in the book.
   verses Verse[]
```
arrocke: I added comments to describe each table and field. Comments with three slashes can later be used by a prisma generator to generate docs for our schema. For now I think linking to the schema file in our markdown docs should be sufficient.

Pertempto:
For some reason, the seed command is failing

```
(base) PS C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation> nx seed api

> nx run api:seed

undefined:1
'{module:commonjs}'
    at parseArgv (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\ts-node\dist\bin.js:69:12)
    at main (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\ts-node\dist\bin.js:25:18)
    at Object.<anonymous> (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\ts-node\dist\bin.js:579:5)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
Node.js v18.14.2
 >  NX   ERROR: Something went wrong in run-commands - Command failed: npx ts-node -O '{"module":"commonjs"}' --project packages/api/tsconfig.json packages/api/prisma/seed.ts
   Pass --verbose to see the stacktrace.

 ————————————————————————————————————————————————————————————————————————————————————————————————————————————————

 >  NX   Ran target seed for project api (2s)

    ×    1/1 failed
    √    0/1 succeeded [0 read from cache]
```
I'll try to figure it out, unless you know off the top of your head what would be causing this. Are you on Windows or some other OS?

Pertempto:
> I'll try to figure it out, unless you know off the top of your head what would be causing this. Are you on Windows or some other OS?

I tried with Powershell, Windows CMD, and GitBash. (I thought maybe it had to do with how PowerShell was interpreting the string). All three complained of JSON syntax error.

Pertempto:
```
"seed": {
    "executor": "nx:run-commands",
    "options": {
    "command": "npx ts-node -O {\\\"module\\\":\\\"commonjs\\\"} --project packages/api/tsconfig.json packages/api/prisma/seed.ts"
    }
},
```
This seemed to work for me. My terminal didn't throw an error so I assume it's running the seed script.

Pertempto:
@arrocke If I pushed a commit with [these changes](https://github.com/arrocke/gloss-translation/pull/49#issuecomment-1520938744), could you see if it works on your end?

arrocke:
I tried running that command on my machine and it didn't work, but I found a better way to solve this. You can specify compiler options for ts-node in tsconfig.json that are only applied with ts-node so that you can configure it distinct from tsc.

Pertempto:
> I tried running that command on my machine and it didn't work, but I found a better way to solve this. You can specify compiler options for ts-node in tsconfig.json that are only applied with ts-node so that you can configure it distinct from tsc.

Ok, great!

Pertempto (approved):
The new seed command works well. 👍🏽

PR #50, opened 2023-04-23
docs: api design for glosses

### LINKED ISSUE #33: api/db design for glosses
## acceptance criteria
* sketch out API routes for the following
  * GET glosses for a verse
  * PATCH gloss for a word
* sketch out db model to support these routes

arrocke:
closes #33


packages/api/public/spec/schemas/parameters.yml
```diff
@@ -0,0 +1,24 @@
+languageCode:
+  name: code
+  in: path
+  description: IETF code of the language
+  required: true
+  schema:
+    type: string
+  example: es
```
arrocke, line 8: I extracted the URL parameters to a common file since they are repeated in multiple URLs.

arrocke (commented):
I've tried to break this down into data that is static and data that is language specific. So the hebrew/greek words, their strongs numbers, and the grammar information is the same regardless of language. So that is in a separate request. `/verses/:verseId/words`

Then for each language, a verse is going to have a natural translation for reference, and a list of glosses for each word. `/language/:code/verses/:verseId/(translation|glosses)`.

Finally to update a gloss, we can use a separate route for words `/language/:code/words/:wordId`.

Pertempto:
> I've tried to break this down into data that is static and data that is language specific. So the hebrew/greek words, their strongs numbers, and the grammar information is the same regardless of language. So that is in a separate request. `/verses/:verseId/words`
> 
> Then for each language, a verse is going to have a natural translation for reference, and a list of glosses for each word. `/language/:code/verses/:verseId/(translation|glosses)`.
> 
> Finally to update a gloss, we can use a separate route for words `/language/:code/words/:wordId`.

I was concerned that we weren't being consistent about plurals in our routes, but I checked the codebase and it looks like we're using `languages`, so we're all good. 😁

packages/api/public/spec/paths/language.yml
Pertempto, line 1: This would've been from your previous PRs, but why have a separate `language.yml`? It seems a little more intuitive to me to have them combined in `languages.yml`. In the future, we're not going to want to have to remember which routes are stored in which file.

packages/api/public/spec/index.yml
```diff
@@ -11,15 +11,15 @@ servers:
...
     $ref: ./paths/language.yml
-  /books/{bookName}/chapters/{chapterNumber}/verses/{verseNumber}/words:
-    $ref: ./paths/verses/words.yml
-  /books/{bookName}/chapters/{chapterNumber}/verses/{verseNumber}/translations/{lang}:
-    $ref: ./paths/verses/translations.yml
-components:
-  schemas:
-    Language:
-      $ref: ./schemas/language.yml
+  /languages/{code}/verses/{verseId}/translation:
```
Pertempto, line 20: The `verseId` system makes the route SO much simpler than it was before! Great idea!!

packages/api/public/spec/paths/language.yml
Pertempto: Ah, I think I might have figured out the answer. It has to do with how OpenAPI connects the routes in `index.yml` with `$ref`, right? `languages.yml` is for the language routes that contain no ID in the path, `language.yml` is for the languages routes that DO contain an ID in the path.

arrocke: Yep, that's the key. The $ref system is a bit clunky in my opinion. To be honest, I'm not super impressed with the OpenAPI schema and tooling. It seems like the good tools around OpenAPI are all paid. However it is good enough for what it needs to do, and I don't want to spend more time looking for something else.

arrocke:
Is this good to merge?

packages/api/public/spec/paths/language.yml
Pertempto: Alright, you could be right. I'm not sure if there really are very great alternatives out there.

Pertempto:
> Is this good to merge?

I had a few more things to check/comment, then I got called away yesterday evening.

packages/api/public/spec/schemas/parameters.yml
```diff
@@ -0,0 +1,24 @@
...
+verseId:
+  name: verseId
+  in: path
+  description: The verse number within the chapter. In the format of `BBCCCVVV`.
```
Pertempto:
```suggestion
  description: The verse number within the chapter. In the format of `BBCCCVVV` (**B**ook, **C**hapter, **V**erse).
```
Just to be very explicit.

packages/api/public/spec/schemas/parameters.yml
```diff
@@ -0,0 +1,24 @@
...
+  required: true
+  schema:
+    type: number
+  example: 0201401
+wordId:
+  name: wordId
+  in: path
+  description: The word ID within the chapter. In the format of `BBCCCVVVWW`.
```
Pertempto:
```suggestion
  description: The word ID within the chapter. In the format of `BBCCCVVVWW` (**B**ook, **C**hapter, **V**erse, **W**ord).
```

Pertempto (changes requested):
I think it would be good to be extra explicit about how the `verseId` and `wordId` patterns work, but otherwise this looks great

packages/api/public/spec/schemas/parameters.yml
Pertempto: (I did test this in my local environment, swagger renders the `**Bold**` Markdown fine)

arrocke: Good idea. That's a concise way to clarify

packages/api/public/spec/paths/language-words.yml
```diff
@@ -0,0 +1,19 @@
+patch:
+  summary: Update the gloss of a word for a language.
+  parameters:
+    - $ref: '../schemas/parameters.yml#languageCode'
+    - $ref: '../schemas/parameters.yml#verseId'
```
Pertempto:
```suggestion
    - $ref: '../schemas/parameters.yml#wordId'
```

Pertempto (commented):
Found this mismatch, otherwise I think we're good.

Pertempto (approved):
I think we're good to merge now.

PR #51, opened 2023-04-26
feat: readonly display line

### LINKED ISSUE #27: readonly display line
## acceptance criteria
- show the hebrew and reference gloss for the selected verse

arrocke:
closes #27


packages/api/prisma/schema.prisma
```diff
@@ -98,14 +114,17 @@ model VerseTranslation {
...
+  wordId     String
+  /// The language of the gloss.
+  language   Language @relation(fields: [languageId], references: [id])
+  /// The ID of the language of the gloss.
+  languageId String   @db.Uuid
+  /// The text of the gloss.
+  gloss      String
+
+  @@id([wordId, languageId])
 }
```
arrocke, line 130: Added a new table to track glosses for words in each language.

packages/api/prisma/seed.ts
```diff
@@ -119,6 +121,21 @@ async function run() {
...
+    },
+  });
+
+  await client.gloss.createMany({
+    data: wordData.map((word) => ({
+      wordId: word.id,
+      languageId: language.id,
+      gloss: word.english,
+    })),
+  });
```
arrocke, line 137: This seed script now also inserts English glosses to use as the default reference gloss

packages/api/public/spec/index.yml
```diff
@@ -6,13 +6,10 @@ info:
     The frontend can be accessed at [https://gloss-translation.vercel.app]().
   version: 0.0.1
 servers:
-  - url: https://gloss-translation-api.vercel.app/api
-    description: Production server
-  - url: http://localhost:4300/api
-    description: Local server
```
arrocke, line 12: I remove this in favor of just a path. By default the explorer will use the URL of the server it is running on. I was getting annoyed with forgetting to change to the local server when testing

packages/api/public/spec/index.yml
```diff
@@ -6,13 +6,10 @@ info:
...
+  - url: /api
 paths:
-  /verses/{verseId}/words:
-    $ref: ./paths/words.yml
+  /verses/{verseId}:
+    $ref: ./paths/verse.yml
```
arrocke, line 12: Instead of a words subresource, I think it makes more sense to just get the verse itself. For now that's just words, but there could be more in the future.

packages/web/src/features/translation/router.tsx
```diff
@@ -1,9 +1,12 @@
...
+import { RouteObject } from 'react-router-dom';
+import TranslationView, { translationViewLoader } from './TranslationView';
 
-const routes: RouteObject[] = [{
-  path: 'translation',
-  element: <TranslationView />
-}]
+const routes: RouteObject[] = [
+  {
+    path: 'translate/:verseId',
```
arrocke, line 6: For now, the verse id is in the URL. That could change with the verse picker

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -1,11 +1,43 @@
...
   const { language } = useLayoutContext();
 
   return (
-    <div className="absolute w-full h-full flex items-center justify-center">
-      selected language: {language?.name ?? 'None'}
+    <div className="px-4">
+      <ol className={`flex flex-wrap ${hebrew ? 'flex-row-reverse' : ''}`}>
+        {verse.data.words.map((word, i) => (
+          <li key={word.id} className="mx-2 mb-4">
+            <div className={`font-serif ${hebrew ? 'text-2xl' : 'text-lg'}`}>
```
arrocke, line 34: We have to modify some of these styles when rendering hebrew in order to get to render right to left. We also need a larger font with Hebrew since they tend to be smaller fonts.

packages/api/prisma/seed.ts
Pertempto: I wondered if we would use that data. 👍🏽

data/seed.dump
Pertempto, line 1: What is the purpose of this file?

Pertempto:
Code looks great, but I'm having troubles running it.  
Accessing the frontend gives me this error:  
image
My route was `http://localhost:4200/translation`. Do I have to have a verse ID? If so, then the root route needs to be changed to not redirect to `/translation`.  
I also see this in the backend logs:
```
API resolved without sending a response for /api/languages, this may result in stalled requests.
```

data/seed.dump
arrocke: To run the seed script on a remote database turns out to take multiple hours because each command is being sent individually over the network rather than a single script file. I used `pg_dump` to export the db data as an archive and then `pg_restore` to send that data to the preview database.

I can either document how to use `pg_restore` with that file or we can replace the `nx seed api` command to use that file instead of the script. If we replace the seed command, we will need to add a command to generate that file when our seed data changes.

arrocke:
I fixed the redirect if you go to the home page. You should see something now

Pertempto:
```
API resolved without sending a response for /api/languages/en/verses/01001002/glosses, this may result in stalled requests.
API resolved without sending a response for /api/verses/01001003, this may result in stalled requests.
API resolved without sending a response for /api/languages, this may result in stalled requests.
API resolved without sending a response for /api/languages/en/verses/01001003/glosses, this may result in stalled requests.
API resolved without sending a response for /api/verses/01001004, this may result in stalled requests.
API resolved without sending a response for /api/languages, this may result in stalled requests.
API resolved without sending a response for /api/languages/en/verses/01001004/glosses, this may result in stalled requests.
API resolved without sending a response for /api/verses/20001004, this may result in stalled requests.
API resolved without sending a response for /api/languages, this may result in stalled requests.
API resolved without sending a response for /api/languages/en/verses/20001004/glosses, this may result in stalled requests.
```
Not sure if these are normal or not.

Pertempto (approved):

packages/api/shared/Route.ts
```diff
@@ -200,10 +200,10 @@ export default function createRoute<Params>(): RouteBuilder<Params> {
       return this;
     },
     build() {
-      return (req: NextApiRequest, res: NextApiResponse) => {
+      return async (req: NextApiRequest, res: NextApiResponse) => {
         const handler = handlers[req.method ?? ''];
         if (handler) {
-          handler(req, res);
+          await handler(req, res);
```
arrocke, line 206: This was the issue with those logs. We have to return a promise from a nextjs handler or next doesn't know that we are still working to send a response

PR #52, opened 2023-04-30
fix: nahum data

Pertempto:
The `bookFiles` array has a typo in the Nahum entry, causing the Nahum data to not be updated correctly. We might need to generate a new `seed.dump` file?

### DIFF
--- a/data/morphology.ts
+++ b/data/morphology.ts
@@ -32,7 +32,7 @@ const bookFiles = [
   'Oba.json',
   'Jon.json',
   'Mic.json',
-  'Neh.json',
+  'Nam.json',
   'Hab.json',
   'Zep.json',
   'Hag.json',


arrocke:
Shoot. Good catch. The command to generate the seed file is in db docs. At this point it might be easier to just reset the preview and production databases with that new seed rather than trying to remove the existing nahum data and replace it

arrocke (approved):

Pertempto:
I updated `seed.dump` using that command. It seems suspicious to me that the file size would have gone up that much.    

image

arrocke:
Yeah, that's not right. Let me take a look later today before we merge

arrocke (approved):
Ok, I ran the pg_dump command and the size went down, which expected because Nehemiah is longer than Nahum. This is the command that I ran:
```
pg_dump --data-only --exclude-table _prisma_migrations -Fc translation > data/seed.dump
```

Pertempto:
> Ok, I ran the pg_dump command and the size went down, which expected because Nehemiah is longer than Nahum. This is the command that I ran:
> 
> ```
> pg_dump --data-only --exclude-table _prisma_migrations -Fc translation > data/seed.dump
> ```


So the command in `db.md` is incorrect? It says to include the db-name.

arrocke:
`translation` is the name of my database

PR #54, opened 2023-05-01
feat: verse selector

### LINKED ISSUE #25: select verse to translate
## acceptance criteria
- type verse in, including abbreviations
- buttons to navigate back and forward one verse
- support book names and shortcuts in each language

Pertempto:
Closes #25   
* Added the `VerseSelector` component, with basic text input and previous/next button navigation.  
    * See `decrementVerseId` and `incrementVerseId` functions for how the previous/next button logic works. It should handle chapter and book boundaries well. I decided to wrap from Revelations 22:21 to Genesis 1:1 (and vice versa), instead of implementing special logic that disables the previous or next buttons for those verses.
* Added a bunch of verse-related helper functions in `verse-utils.ts`. It seems like some of these potentially could be useful in the backend codebase at some point. Could we have shared utils?  
* See `book-terms/README.md` for a simple explanation of how that system works.


Pertempto:
> The language here is the language being translated into, not the interface language. So to render the reference we want to use i18next and useTranslation. We can create a separate translation namespace called bible.json for all of the Bible specific strings. Then you can use t to convert the verse id in a reference in the language of the interface.  

I'm not sure if the translation namespace scheme will work? The `book-terms/{langCode}.json` files each contain an array with an array for each book, and the book arrays are variable-length depending on how many different terms/abbreviations apply to each book. I just use these terms when parsing the reference from the user. The only way they are used for output is that I use the first item in each array as the book title. See  `book-terms/README.md` and `book-terms/en.json`.

Pertempto:
> This is a good idea. In addition, the strings used for lookup should probably come from translation strings as well. i18next has [methods to get translation strings](https://www.i18next.com/overview/api#resource-handling) that might be useful to build something that can parse a reference string in the interface language.  


Good point. Currently I'm just using a hard coded format string like `bookName chapterNumber:verseNumber`, but that should come from translation strings.

Pertempto:
I updated my reference generation code to use `t()`. I forgot to update my parsing regex. Not sure how we want to handle that.

packages/web/src/shared/utils.ts
```diff
@@ -4,3 +4,14 @@ export function capitalize(str: string) {
...
 }
+
+/**
+ * Bound a number into the inclusive range [min, max].
+ * @param num The input number.
+ * @param min The minimum bound.
+ * @param max The maximum bound.
+ * @returns A number in the inclusive range [min, max].
+ */
+export function bound(num: number, min: number, max: number) {
```
arrocke: Can we call this clamp? When I've seen this kind of function before, that's what it is called.

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Icon } from './Icon';
+import TextInput from './TextInput';
+
+
+export interface VerseSelectorProps {
+  verseId: string;
+  goToVerse: (verseId: string) => any;
+}
+
+export function VerseSelector({ verseId, goToVerse }: VerseSelectorProps) {
```
arrocke: This should be a component in `features/translation` since it is specific to that view

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+        goToVerse(newVerseId);
+      }
+    }
+  }
+
+  return (
+    <div className='flex flex-row gap-4 items-center'>
+      <TextInput autoComplete="off" placeholder={reference} onChange={onInputChange} onKeyDown={onKeyDown} />
+      <button onClick={() => goToVerse(decrementVerseId(verseId))}>
+        <Icon icon="arrow-left" />
```
arrocke: We should add a span with the `sr-only` class with the text `Next Verse` and `Previous Verse` to each of these buttons. That will render text for screen readers so that they know what these buttons are.

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+  verseId: string;
+  goToVerse: (verseId: string) => any;
+}
+
+export function VerseSelector({ verseId, goToVerse }: VerseSelectorProps) {
+  const { t, i18n } = useTranslation();
+  const langCode = i18n.language;
+  const verseInfo = parseVerseId(verseId);
+  const reference = t('reference_format', { ...verseInfo, bookName: bookName(verseInfo.bookId, langCode) });
+  const [newReference, setNewReference] = useState('');
```
arrocke: Rather than managing this state here, we can just get the value from the event when we need it. `e.currentTarget.value`

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  const onInputChange = (e: any) => {
+    e.preventDefault();
+    setNewReference(e.target.value);
+  }
+
+  const onKeyDown = (e: any) => {
+    if (e.keyCode == 13) {
```
arrocke: We can use `e.key === 'Enter'` for this. `keyCode` is deprecated. We can also type the keydown event by import `KeyboardEvent` from React. and typing this `KeyboardEvent<HTMLInputElement>`

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { bookName, decrementVerseId, incrementVerseId, parseReference, parseVerseId } from '../verse-utils';
+import { Icon } from './Icon';
+import TextInput from './TextInput';
+
+
+export interface VerseSelectorProps {
+  verseId: string;
+  goToVerse: (verseId: string) => any;
```
arrocke: The react convention for function props is `onVerseChange`. Almost all events from components are `onEventName`

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const [newReference, setNewReference] = useState('');
+
+  const onInputChange = (e: any) => {
+    e.preventDefault();
+    setNewReference(e.target.value);
+  }
+
+  const onKeyDown = (e: any) => {
+    if (e.keyCode == 13) {
+      e.preventDefault();
```
arrocke: I don't think we need prevent default on these event handlers

packages/web/src/shared/components/VerseSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+export interface VerseSelectorProps {
+  verseId: string;
+  goToVerse: (verseId: string) => any;
+}
+
+export function VerseSelector({ verseId, goToVerse }: VerseSelectorProps) {
+  const { t, i18n } = useTranslation();
+  const langCode = i18n.language;
+  const verseInfo = parseVerseId(verseId);
+  const reference = t('reference_format', { ...verseInfo, bookName: bookName(verseInfo.bookId, langCode) });
```
arrocke: Do we know if the verse format changes from language to language or just the book names? If not we can do something like this in the JSX below:
```
{t(bookCode)} {chapterNumber}:{verseNumber}
```

packages/web/src/shared/verse-utils.ts
```diff
@@ -0,0 +1,161 @@
+// This file contains utility functions related to verse IDs.
+
+import verseCounts from '../assets/verse-counts.json';
```
arrocke: I think both of this file and the verse counts json can be moved to the features/translation directory so they are together

packages/web/src/shared/verse-utils.ts
```diff
@@ -0,0 +1,161 @@
...
+  return bookTerms[bookId - 1][0];
+}
+
+/**
+ * Try to parse a verse ID from a reference, in the given language.
+ * @param reference The string that should be parsed. In the form "bookName chapterNumber:verseNumber". Example: "Exo 3:14".
+ * @param langCode The language to use.
+ * @returns The verse ID if it can be determined, otherwise `null`.
+ */
+export function parseReference(reference: string, langCode: string): string | null {
```
arrocke: I'm thinking we set up the `bible.json` localization file to have these keys for each language:
```
// use these two for building strings
gen: 'Genesis',
gen_short: 'Gen',
// use these two for parsing references
gen_match: ['Genesis', 'Gen', 'Gn']
```

We might need to move this into the verse selector component so that we can use `t('gen', { context: 'match' })` to get the translation strings for the language.

Doing it this way allows `i18next` to load the translation strings at runtime rather than bundling them all into the code.

packages/web/src/assets/book-terms/README.md
```diff
@@ -0,0 +1,21 @@
+# Overview
```
arrocke: All of these notes, we will want to move to the localization docs

arrocke (commented):
If it would help to work together on some of this, let me know.

packages/web/src/shared/utils.ts
Pertempto: Sure! I've only seen it once as bound, but clamp is a natural name.

packages/web/src/shared/components/VerseSelector.tsx
Pertempto: Ok. The reason I put it here was that I thought we might use it other places in the future. I guess if that does happen we can  move it at that time.

packages/web/src/shared/components/VerseSelector.tsx
arrocke: Given the conversation in whatsapp, what you have here is what we want

packages/web/src/shared/verse-utils.ts
Pertempto: Alright! Next time could you tell me things like this earlier? I posted my plan here https://github.com/arrocke/gloss-translation/issues/25#issuecomment-1512108308 and never got any feedback that we would want a different system.

packages/web/src/assets/book-terms/README.md
Pertempto: Note for self: still needs done

packages/web/src/shared/verse-utils.ts
Pertempto: Note for self: still needs done

packages/web/src/shared/components/VerseSelector.tsx
Pertempto: Note for self: still needs done

Pertempto:
@arrocke The way I'm used to doing code review, we let the reviewer resolve the threads so that they can be sure they got to double check that everything actually got handled correctly. If you prefer a different strategy, let me know.  
I have a few more threads to address, hopefully I can handle them all Monday.

packages/web/src/shared/verse-utils.ts
arrocke: > Alright! Next time could you tell me things like this earlier? I posted my plan here https://github.com/arrocke/gloss-translation/issues/25#issuecomment-1512108308 and never got any feedback that we would want a different system.

That comment was a summary from our discussion, so at the time I thought that was the best approach. It didn't occur to me that parsing user input of a verse was going to need to be multilingual until you asked the question about formatting the verse reference in other languages.

arrocke:
That is my preference as well. This is probably something we will want to document in our contributor documentation

packages/web/src/shared/verse-utils.ts
Pertempto: Thanks for the clarification!

packages/web/src/shared/components/VerseSelector.tsx
Pertempto: Done

packages/web/src/shared/verse-utils.ts
Pertempto: Done

packages/web/src/assets/book-terms/README.md
Pertempto: Moved

Pertempto:
@arrocke This should be ready again! Hopefully I didn't miss any of your suggestions. I only have `bible.json` for `en`, not `es`.  
There are many other aliases we might want to add in the future. I think it might be useful to implement some kind of fuzzy match algorithm instead of trying to store every possible alias for each book in each language.

packages/web/src/features/translation/VerseSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function VerseSelector({ verseId, onVerseChange }: VerseSelectorProps) {
+  const { t } = useTranslation(["translation", "bible"]);
+  const verseInfo = parseVerseId(verseId);
+
+  const onKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key == "Enter") {
+      const newReference = (e.target as HTMLInputElement).value;
+      (e.target as HTMLInputElement).value = "";
```
arrocke: If we use `currentTarget`, then we don't need to cast the type of target. In this case currentTarget and target are the same because there are no nested elements inside the input

packages/web/src/features/translation/VerseSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+  decrementVerseId,
+  generateReference,
+  incrementVerseId,
+  parseReference,
+  parseVerseId
+} from "./verse-utils";
+
+export interface VerseSelectorProps {
+  verseId: string;
+  onVerseChange: (verseId: string) => any;
```
arrocke: We can type the return type as `void` because we don't expect anything to be returned

packages/web/src/features/translation/VerseSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }
+    }
+  };
+
+  return (
+    <div className="flex flex-row gap-4 items-center">
+      <TextInput
+        autoComplete="off"
+        placeholder={generateReference(verseInfo, t)}
+        onKeyDown={onKeyDown}
```
arrocke, line 41: Let's add `aria-label="select verse"` for screen readers

packages/web/src/public/locales/en/bible.json
```diff
@@ -0,0 +1,250 @@
+{
+  "gen": "Genesis",
+  "gen_short": "Gen",
+  "gen_match": ["Genesis", "Gen", "Gn"],
```
arrocke, line 6: I think you are absolutely right about using an algorithm to deal with all the possible variations. Generating all of these is going to be tedious and error prone. Let's tackle that as a separate story

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -72,67 +75,71 @@ export default function TranslationView() {
...
+  const isHebrew = bookId < 40;
   return (
     <div className="px-4">
-      <ol className={`flex flex-wrap ${isHebrew ? 'flex-row-reverse' : ''}`}>
-        {verse.words.map((word, i) => {
-          const targetGloss = targetGlosses[i]?.gloss;
-          const isSaving = glossRequests.some(
-            ({ wordId }) => wordId === word.id
-          );
+      <div className='flex flex-col gap-2'>
```
arrocke: How did you decide to add this extra div rather than putting these classes on the parent div?

packages/web/src/features/translation/verse-utils.ts
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Try to parse a verse ID from a reference, using the i18n translation function.
+ * @param reference The string that should be parsed. In the form "bookName chapterNumber:verseNumber". Example: "Exo 3:14".
+ * @param t The i18n translation function to use.
+ * @returns The verse ID if it can be determined, otherwise `null`.
+ */
+export function parseReference(reference: string, t: TFunction): string | null {
+  // Parse the reference into three parts.
+  const referenceRegex = new RegExp(
+    t("reference_regex", { ns: "translation" }) as string
```
arrocke: Should the namespace be `bible`?

arrocke (changes requested):
A couple small things and a few questions. I'm really excited by how flexible this is

packages/web/src/features/translation/VerseSelector.tsx
Pertempto: I had to cast, otherwise I get errors.  
image

packages/web/src/features/translation/VerseSelector.tsx
Pertempto: We want that to be translated, right?

Pertempto: I used `arial-label={t('select_verse')}`

packages/web/src/features/translation/TranslationView.tsx
Pertempto: Hmmm, not sure now. Maybe I did it after merge conflicts.

packages/web/src/features/translation/verse-utils.ts
Pertempto: Good point! I had been thinking that `bible` would be only for book names and terms, but really it makes sense for the regex and format string to be in that file too. It's all related to reference parsing.

Pertempto:
I think I addressed all change requests.

packages/web/src/features/translation/verse-utils.ts
Pertempto: I also updated `i18n.md`. In the future we'll want to look at labeling the groups in the `reference_regex` (in case the order changes between languages), and look into flexible number formatting in `reference_format`.

arrocke (approved):
Nice work!

arrocke:
I added #57 to address the algorithmic approach. In response to your comment about labelling the capture groups of the regex, I think we might be able to generate the regex from the format string. We just need to swap out each of the template properties for capture groups.

Pertempto:
> In response to your comment about labelling the capture groups of the regex, I think we might be able to generate the regex from the format string. We just need to swap out each of the template properties for capture groups.

Yeah I think we probably could. I was just done messing with the logic in this issue, we can fine tune it in future issues

PR #58, opened 2023-05-10
feat: select from list of previous glosses

### LINKED ISSUE #29: track glosses for each lemma
## acceptance criteria
- keep track of the glosses used for each form of each lemma
- show list of glosses in dropdown on the gloss input for each word sorted by use

arrocke:
closes #29 

* Adds TypeaheadInput component
* Adds queries to see the list of glosses for a word in a single language
* Updates the translation view to show dropdown of glosses to pick from.


package.json
```diff
@@ -39,6 +39,7 @@
     "@typescript-eslint/eslint-plugin": "5.36.1",
     "@typescript-eslint/parser": "5.36.1",
     "autoprefixer": "10.4.13",
+    "downshift": "7.6.0",
```
arrocke, line 42: downshift is great for building accessible and easy to use autocomplete/typeahead components

packages/api/prisma/schema.prisma
```diff
@@ -124,7 +124,7 @@ model Gloss {
   /// The ID of the language of the gloss.
   languageId String   @db.Uuid
   /// The text of the gloss.
-  gloss      String
```
arrocke, line 127: the gloss needs to be optional in the case that it is removed

packages/web/src/shared/components/TypeaheadInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export interface TypeaheadInputProps {
+  className?: string;
+  value?: string;
+  labelId?: string;
+  items: TypeaheadInputItem[];
+  onChange(value?: string): void;
+}
+
+export default function TypeaheadInput({
```
arrocke: This input is meant to provide autocomplete from a predefined list. As you type it will narrow suggestions down and provide an option to create a new element

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -0,0 +1,69 @@
...
+  word: { id: string; text: string };
+  referenceGloss?: string;
+  gloss?: string;
+  previousGlosses: string[];
+  originalLanguage: 'hebrew' | 'greek';
+  status: 'empty' | 'saving' | 'saved';
+  onGlossChange(gloss?: string): void;
+}
+
+export default function TranslateWord({
```
arrocke, line 17: This has everything that was in the view for each word before, but I swapped the text input for the new typeahead

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -32,34 +30,58 @@ export default function TranslationView() {
...
-    onMutate: ({ wordId }) => {
+    onMutate: async ({ wordId, gloss }) => {
       const requestId = Math.floor(Math.random() * 1000000);
       setGlossRequests((requests) => [...requests, { wordId, requestId }]);
-      return { requestId };
+
+      const queryKey = ['verse-glosses', language, params.verseId];
+      await queryClient.cancelQueries({ queryKey });
+      const previousGlosses = queryClient.getQueryData(queryKey);
+      queryClient.setQueryData<GetVerseGlossesResponseBody>(queryKey, (old) => {
```
arrocke, line 42: This mutation to update glosses is a little more complex now. When we fire the mutation we optimistically update the cache with the new gloss.

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -32,34 +30,58 @@ export default function TranslationView() {
...
+        context?.previousGlosses
+      );
+
       alert('Unknown error occurred.');
     },
-    onSuccess: (_, { wordId, gloss }, context) => {
+    onSettled: (_, __, ___, context) => {
+      queryClient.invalidateQueries({
+        queryKey: ['verse-glosses', language, params.verseId],
+      });
```
arrocke, line 78: When the mutation finishes, then we instruct react-query to refetch that query to make sure it is up to date

arrocke:
One thing we are going to need to handle in the future is how to handle languages with diacritics. That is going to implicate how we generate the list of unique glosses and how we match in the dropdown.

packages/web/src/shared/components/TypeaheadInput.tsx
Pertempto: Why not have a brief comment above the component that explains this?

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -0,0 +1,74 @@
...
+          }
+        }}
+      />
+      <InputHelpText id={`word-help-${word.id}`}>
+        {(() => {
+          if (status === 'saving') {
+            return (
+              <>
+                <Icon icon="arrows-rotate" className="mr-1" />
+                Saving...
```
Pertempto: This and line 64 should be i18n-ized.

packages/web/src/shared/components/TypeaheadInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Text input with autocomplete suggestions.
+ *
+ * Suggestions have the shape of `{ label: string; value: string; }`.
+ * When an item is selected, its value will be passed to the `onChange` event.
+ *
+ * To support the creation of new items, attach the `onCreate` event.
+ * The text value will be passed to the event.
+ */
+export default function TypeaheadInput({
```
Pertempto:
```suggestion
export default function TypeAheadInput({
```
Are you wanting "type ahead" to be one word?

Pertempto: image

Something isn't quite right with z-index it seems.

Pertempto (changes requested):
Overall works pretty well. The horizontal scrollbar on the autocomplete doesn't look very good. How hard would it be to remove it?  

I can't really approve until the UI glitch is fixed.

packages/web/src/shared/components/TypeaheadInput.tsx
arrocke: I decided to rename it AutocompleteInput, which I think it is a lot less ambiguous

Pertempto: But why don't you capitalize "Complete" to make it `AutoCompleteInput`?

arrocke: I see it as one word every where I've seen it

Pertempto: 😬 my bad

Pertempto (approved):
It looks a lot better without the extra scrollbar!

PR #90, opened 2023-05-20
feat: variable width word columns

### LINKED ISSUE #55: make word columns variable-width
The columns for each word in the original language should be variable width. They look way too big for short words, and don't handle long glosses in an ideal way either.  

image

Pertempto:
Closes #55 

I really don't like the method behind this, but I couldn't figure out anything that would work with just plain styling.  
For some reason, the `TextInput` inside `AutocompleteInput` would never behave, if I tried to have the `AutocompleteInput` implicitly get it's width from the invisible measurement text element.


packages/web/src/shared/components/AutocompleteInput.tsx
```diff
@@ -26,7 +26,6 @@ export interface AutocompleteInputProps
  * The text value will be passed to the event.
  */
 export default function AutocompleteInput({
-  className = '',
```
Pertempto: I didn't see this being used anywhere, so I removed it.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -26,16 +26,16 @@ export default function TranslateWord({
   const { t } = useTranslation();
 
   return (
-    <li className="mx-2 mb-4 w-36">
+    <li className="mx-2 mb-4">
       <div
         id={`word-${word.id}`}
-        className={`font-serif mb-2 ${
-          originalLanguage === 'hebrew' ? 'text-2xl text-right' : 'text-lg'
+        className={`font-serif mb-2 text-center ${
```
arrocke: I'm not generally a fan of of center alignment. It is a general principle in UX design to align the edges of everything. But if the problem is that having a RTL and a LTR language on top of each other also looks weird, I would solve that problem by aligning everything to the right for Hebrew and to the left for Greek. That's what the current app does and I think it works well.

packages/web/src/shared/components/AutocompleteInput.tsx
```diff
@@ -96,13 +100,35 @@ export default function AutocompleteInput({
...
+  // The width of textMeasureElement allows us to measure the width of the input text.
+  // That width is used to size the overall component.
+  const textMeasureElement = useRef(null);
+  const [measuredWidth, setMeasuredWidth] = useState(0);
+  function updateMeasuredWidth() {
+    if (textMeasureElement.current) {
+      setMeasuredWidth((textMeasureElement.current as HTMLElement).clientWidth);
+    }
+  }
+  useEffect(() => updateMeasuredWidth(), [textMeasureElement]);
```
arrocke: What do you think about pulling this out into its own custom hook? We could use it like this in TranslateWord.
```js
const [text, setText] = useState('asdf')
const { width } = useTextWidth(text)

return <AutocompleteInput value={text} style={{ width: width + 56 }} />
```

That hook can also manage its own element to use for measuring:
```js
const el = useRef<HTMLDivElement>()
useEffect(() => {
  const div = document.createElement('div')
  document.body.appendChild(div)
  el.current = div
  return () => div.remove()
}, [])

useEffect(() => {
  el.current.innerText = text
}, [text])
```

Pertempto: I like this idea! I'll try to implement it tonight.

packages/web/src/features/translation/TranslateWord.tsx
Pertempto: I had never heard of that principle, but it makes sense. It looks better like you suggested.

Pertempto: Aha! https://uxengineer.com/principles-of-design/alignment/

Pertempto:
I've implemented the `useTextWidth` hook. It seems to work well.  
If you type a long gloss into text box that was previously narrow width, the experience isn't very great. I think we can improve that in the future when we have actual users to give us feedback.  

Occasionally `AutocompleteText` gets in a weird state where it has a new gloss, but the `onChange` callback is not called. It seems to happen when there are no autocomplete suggestions in the list, but the user tabs out of the input without selecting the "create" item. I'm not sure if this is a bug or not. Whatever it is, it doesn't feel like a smooth experience, and should receive some attention in the future.

arrocke (approved):
> If you type a long gloss into text box that was previously narrow width, the experience isn't very great. I think we can improve that in the future when we have actual users to give us feedback.

I added a commit that I think improves this. Let me know what you think. This is what it does:
* Set a min width on the autocomplete itself
* Let the options in the autocomplete expand within reason

> Occasionally AutocompleteText gets in a weird state where it has a new gloss, but the onChange callback is not called.

In the flash message PR, I added `@headless-ui/react` which has a bunch of useful complex components that are designed to work with custom tailwind styles. They have an autocomplete that we could use instead of downshift. Let's tackle any remaining usability issues when we do that.

Pertempto:
Somehow merging in main broke this. I wonder if it has to do with your form changes.  
```
Warning: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
```

Pertempto:
> Somehow merging in main broke this. I wonder if it has to do with your form changes.
> 
> ```
> Warning: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
> ```

It seems to be triggered by `SelectInput` and `TextInput`

arrocke:
That should be fixed in in [#92](https://github.com/arrocke/gloss-translation/pull/92/files#diff-55c9078b77d2dd56ad86070cf5cd8483c9f73d1d7e51eb39c003d297fbfcf7ed). I didn't quite get the Select and TextInputs right so that they can work both inside and outside a form

Pertempto:
> That should be fixed in in [#92](https://github.com/arrocke/gloss-translation/pull/92/files#diff-55c9078b77d2dd56ad86070cf5cd8483c9f73d1d7e51eb39c003d297fbfcf7ed). I didn't quite get the Select and TextInputs right so that they can work both inside and outside a form

So you're fine with me merging this, even though things will be broken?

arrocke:
Yep, go ahead. It's already broken in production. In a few months from now, when we have users, things will be different

PR #92, opened 2023-05-22
feat: flash messages

arrocke:
This was another generic feature in building out user invitations that I thought would be good to pull into a separate PR


package.json
```diff
@@ -27,6 +27,7 @@
     "@fortawesome/fontawesome-svg-core": "6.3.0",
     "@fortawesome/free-solid-svg-icons": "6.3.0",
     "@fortawesome/react-fontawesome": "0.2.0",
+    "@headlessui/react": "1.7.14",
```
arrocke, line 30: This library has a great react component for transition, and great components to use for dropdowns, modals, etc that are designed for tailwind. We can replace some of the other libraries I've added with this one in a future PR.

packages/web/src/shared/components/form/SubmitButton.tsx
arrocke, line 1: This submit button as designed isn't super accessible because the screen reader doesn't know whether to read the changing button text or the flash message. I've replaced it with a spinner to the right, that announces the form is submitting, so that the submit button itself isn't changing. Now when the form finishes, just the flash message is read

packages/web/src/shared/hooks/flash.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+export interface FlashProviderProps {
+  children: ReactNode;
+}
+
+function generateId() {
+  return Math.floor(Math.random() * 100_000_000).toString();
+}
+
+export function FlashProvider({ children }: FlashProviderProps) {
```
arrocke, line 56: This flash provider manages the state and UI for flash messages.

packages/web/src/shared/hooks/flash.tsx
```diff
@@ -0,0 +1,164 @@
...
+  );
+}
+
+export function useFlash() {
+  const context = useContext(FlashContext);
+  if (!context) {
+    throw new Error('useFlash should be used within FlashContext component.');
+  }
+  return context;
+}
```
arrocke, line 193: In order to display a flash message, you can use this hook to get access to the functions in the flash provider

packages/web/src/features/languages/NewLanguageView.tsx
```diff
@@ -28,20 +31,22 @@ export default function NewLanguageView() {
...
           (error) => error.code === 'AlreadyExists'
         );
         if (alreadyExistsError) {
-          alert(`Language with code "${data.code}" already exists.`);
+          flash.error(t('language_exists', { code: data.code }));
           return;
         }
       }
-      throw error;
+      flash.error(`${error}`);
```
arrocke, line 48: Here you can see that in action with translation

package.json
Pertempto: I took a look at it's site. It looks pretty nice!

Pertempto:
Code looks good. I haven't been able to test because of CORS errors with my local server. Not sure why it started happening yesterday.

Pertempto:
I just tried the preview. The CORS error occur there too.

Pertempto (approved):
They look good! This also gave me the chance to learn about contexts in React 😄

PR #94, opened 2023-05-25
fix: cors issue and non form inputs

arrocke:
* Fixes issues with CORS preflight requests
* Fixes issues with input components used outside of a form
* Fixes issues with slashes in branch names of preview apps

### DIFF
--- a/packages/api/shared/Route.ts
+++ b/packages/api/shared/Route.ts
@@ -108,8 +108,6 @@ export default function createRoute<Params>(): RouteBuilder<Params> {
     definition: RouteDefinition<Params, RequestBody, ResponseBody>
   ): NextApiHandler {
     return async (req, res) => {
-      await cors(req, res);
-
       const responseHelper: ResponseHelper<ResponseBody> = {
         ok(body?: ResponseBody) {
           if (body) {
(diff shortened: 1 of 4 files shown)


packages/api/shared/Route.ts
```diff
@@ -205,6 +203,8 @@ export default function createRoute<Params>(): RouteBuilder<Params> {
     },
     build() {
       return async (req: NextApiRequest, res: NextApiResponse) => {
+        await cors(req, res);
```
arrocke, line 206: We have to handle cors here so that it can provide a response for an OPTIONS request. Otherwise this will return 405 because there is no handler for it

packages/web/src/shared/components/form/SelectInput.tsx
```diff
@@ -8,17 +8,15 @@ export interface SelectInputProps
...
     const context = useFormContext();
-    const hasErrors = !!context?.formState.errors[name];
-    const registerProps = context?.register(name, {
+    const hasErrors = !!context?.formState.errors[props.name];
+    const registerProps = context?.register(props.name, {
       required,
-      onChange,
-      onBlur,
+      onChange: props.onChange,
+      onBlur: props.onBlur,
```
arrocke, line 17: This is what broken the SelectInput. I was destructuring props to get the change and blur handlers for the validation library, but when there is no validation these, still need to be attached to the input element

Pertempto (approved):

PR #96, opened 2023-05-26
feat: access control for system admins

### LINKED ISSUE #67: system admin role
## acceptance criteria
* Some users need to be given the role of system administrator
* Users with the system administrator should be the only people who can:
  * Add new people to the platform
  * Create new languages

## notes
* In a future story, language administrators will be able to invite people to a language

arrocke:
closes #67 

* Adds system role that can be granted to users
* Adds basic access control system which grants the ability to create users and languages to admins

To test this, you will need to add the ADMIN role to a user directly in the database.


packages/api/shared/access-control/policy.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  if (user) {
+    can('translate', 'Language');
+    can('read', 'Language');
+
+    if (user.systemRoles.includes(SystemRole.ADMIN)) {
+      can('create', 'Language');
+      can('create', 'User');
+    }
+  }
```
arrocke, line 31: Casl is a great library for managing access control. You can see here that the `can` function specifies the action you want to take on which resource.

Additionally, casl has an adapter to prisma, so we can specify conditions the same as a `where` property on a prisma query:
```ts
can('read', 'Language', { members: { some: { id: user.id } } })
```

packages/api/shared/access-control/authorize.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Determine whether a user is authorized to perform the requested action on the given subject.
+ *
+ * If no subject ID is given, the user will be authorized if they have access to _any_ (not all) of that subject type.
+ * @param options
+ */
+export function authorize<Params, Body>(
+  options:
+    | AuthorizeOptions
+    | ((req: RouteRequest<Params, Body>) => AuthorizeOptions)
+): (req: RouteRequest<Params, Body>) => Promise<void> {
```
arrocke, line 23: This authorize function makes authorization easy for routes. You just need to convert the request into an object with the action, and subject ID. This function will load the correct data to send through the policy and throw Forbidden and NotFound errors as appropriate

packages/api/shared/access-control/authorize.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const policy = createPolicyFor(req.session?.user);
+
+    let subject: Subject | undefined = undefined;
+    if ('subjectId' in config && config.subjectId) {
+      switch (config.subject) {
+        case 'Language': {
+          const language = await client.language.findUnique({
+            where: { code: config.subjectId },
+          });
```
arrocke, line 35: In the future we will want to load language members so that the policy can test against that data as well

packages/api/prisma/schema.prisma
```diff
@@ -17,13 +17,46 @@ datasource db {
...
+  systemRoles   UserSystemRole[]
+  /// The list of sessions for the user.
+  sessions      Session[]
+}
+
+/// A authenticated session for a user.
+model Session {
+  id           String   @id @default(dbgenerated("generate_ulid()")) @db.Uuid
+  /// The session token from the cookie.
+  sessionToken String   @unique
```
arrocke, line 37: I ended up moving back to server sessions instead of JWTs so that if roles change, the JWT isn't out of date

packages/api/pages/api/auth/[...nextauth].ts
```diff
@@ -26,74 +36,88 @@ const ORIGIN_MATCH = process.env.ORIGIN_MATCH
...
-          // To mitigate this, we obfuscate the response time by adding some random delay.
-          await new Promise((resolve) =>
-            setTimeout(resolve, 1800 + Math.floor(Math.random() * 500))
-          );
-          // We redirect to the verify login page rather than showing an error message
-          // so that an attacker can't determine if the account exists.
-          return '/api/auth/verify-request';
-        }
-      },
+        session.user.systemRoles = roles.map(({ role }) => role);
```
Comment (arrocke) [inline packages/api/pages/api/auth/[...nextauth].ts:73]: In order for this to work we have to load the users roles from the database in their session. This will be used both on the server for access control, and by the frontend to hide and show UI

packages/api/pages/api/languages/index.ts
```diff
@@ -21,6 +22,7 @@ export default createRoute<void>()
   })
   .post<PostLanguageRequestBody, void>({
     schema: languageSchema,
+    authorize: authorize({ action: 'create', subject: 'User' }),
```
Pertempto:
```suggestion
    authorize: authorize({ action: 'create', subject: 'Language' }),
```
If I'm getting the casl stuff right, the this should be `Language`. Maybe I'm misunderstanding something.

Pertempto (changes requested):

Pertempto (approved):

PR #99, opened 2023-05-27
feat: users view

### LINKED ISSUE #73: user search
## acceptance criteria
* As a system admin, I can view a list of users and what languages they have access to
* I can give users system admin privileges

arrocke:
closes #73


packages/api/shared/Route.ts
```diff
@@ -153,26 +156,33 @@ export default function createRoute<Params>(): RouteBuilder<Params> {
...
+        const session =
           (await getServerSession(req, res, authOptions)) ?? undefined;
 
-        await definition.authorize?.(retypedReq);
+        const request: RouteRequest<Params, RequestBody> = {
+          query: req.query as Params,
+          body,
+          session,
+          policy: createPolicyFor(session?.user),
+        };
```
arrocke, line 180: I needed a way to make the policy available in the route, so making a new request object rather than using the nextjs request was easier with typescript

packages/api/pages/api/users/index.ts
```diff
@@ -1,10 +1,40 @@
...
+  .get<void, GetUsersResponseBody>({
+    authorize: authorize({
+      action: 'read',
+      subject: 'User',
+    }),
+    async handler(req, res) {
+      const policy = createPolicyFor(req.session?.user);
+
+      const users = await client.user.findMany({
+        where: accessibleBy(policy).User,
```
arrocke: Now that we have the policy in the route, we can write a query that returns only the users the authenticated user has access to.

packages/web/src/features/users/UsersView.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <ListBody>
+            {usersQuery.data?.data.map((user) => (
+              <ListRow key={user.id}>
+                <ListCell header>{user.name}</ListCell>
+                <ListCell>{user.email}</ListCell>
+                <ListCell>
+                  <SelectInput
+                    className="w-32"
+                    name="userRole"
+                    value={user.systemRoles[0] ?? ''}
```
arrocke: Since there is only one role at the moment, a select input works just fine. In a separate PR, I want to replace downshift with headless-ui in the autocomplete component and also add support for multiple selection, for when we add more roles

packages/web/src/public/locales/en/translation.json
Pertempto, line 1: We'll need the new terms in `ar/translation.json` too.

packages/api/pages/api/users/[userId].ts
```diff
@@ -0,0 +1,59 @@
+import * as z from 'zod';
+import { SystemRole, UpdateUserRequestBody } from '@translation/api-types';
```
Comment (Pertempto) [inline packages/api/pages/api/users/[userId].ts:2]: This occurs when I try to update the role of my own user to "Admin".  
```
Module not found: Can't resolve '@translation/api-types'
  1 | import * as z from 'zod';
> 2 | import { SystemRole, UpdateUserRequestBody } from '@translation/api-types';
  3 | import { authorize } from '../../../shared/access-control/authorize';
  4 | import createRoute from '../../../shared/Route';
  5 | import { client } from '../../../shared/db';

https://nextjs.org/docs/messages/module-not-found
```
I'm not sure what's going on, because I can clearly see `@translation/api-types` in `tsconfig.base.json`.

Comment (arrocke) [reply packages/api/pages/api/users/[userId].ts:2]: Can you try running `nx reset` and then run the servers again?

Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:2]: It still occurs. I tried restarting my computer and doing it again, still happened. I tried deleting the `npm_modules/.cache` directory, did it again, and the error still occurs. I'll keep trying next week.

Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:2]: It's strange that it only occurs there, because many other files import `@translation/api-types`.

Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:2]: I tried temporarily to change that line to  `import { SystemRole, UpdateUserRequestBody } from '../../../../api-types/src/index';`. Then I got a 403 error because I don't have permission to update the role of a user. I have to manually set myself as an admin in the DB? How do I do that? I tried and got errors.

Comment (arrocke) [reply packages/api/pages/api/users/[userId].ts:2]: Use this script:
```sql
INSERT INTO "public"."UserSystemRole" ("userId", "role")
VALUES ('<user-id>', 'ADMIN');
```

Comment (arrocke) [reply packages/api/pages/api/users/[userId].ts:2]: So, I was finally able to reproduce it. I traced the issue to the import of SystemRole. For some reason because that is not just a typescript interface it creates an issue

Comment (arrocke) [reply packages/api/pages/api/users/[userId].ts:2]: Ok, I upgraded nx to the latest version, and the issue seemed to go away. Can you try again and confirm?

Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:2]: It worked without a hitch this time. Thanks!

Pertempto (approved):

PR #101, opened 2023-05-30
feat: right to left layout

### LINKED ISSUE #85: right to left layout
## acceptance criteria
* Reverse the layout for languages that are right to left

Pertempto:
To Do
- [x] Add Arabic book names
- [x] Look at replacing `tailwindcss-rtl` with alternative method


package.json
```diff
@@ -65,6 +65,7 @@
     "react-i18next": "12.2.0",
     "react-refresh": "0.10.0",
     "tailwindcss": "3.2.7",
+    "tailwindcss-rtl": "^0.9.0",
```
Pertempto: Yes, this library might not be the best supported, but it does what it does well. I think it is easier to read and write `me-4` than `ltr:ml-4 rtl:mr-4`, especially when we will have that kind of thing all throughout the codebase. The `start-[x]` and `end-[x]`  utilities are also very useful, instead of `ltr:left-[x] rtl:right-[x]`.

The only bug I really ran into was that the library doesn't handle negative numbers. https://github.com/20lives/tailwindcss-rtl/issues/53. It does look like the maintainer has quite releasing updates.

packages/web/src/public/locales/ar/bible.json
Pertempto, line 1: I didn't even try to translate the `short` versions, I didn't know where to start. We definitely will want an actual Arabic speaker to update all the translations in this file at some point.

packages/web/src/public/locales/ar/bible.json
```diff
@@ -0,0 +1,135 @@
...
+  "1sa_short": "1 Sam",
+  "2sa": "صموئيل الثاني 2",
+  "2sa_short": "2 Sam",
+  "1ki": "الملوك الأول 1",
+  "1ki_short": "1 Kgs",
+  "2ki": "لملوك الثاني 2",
+  "2ki_short": "2 Kgs",
+  "1ch": "أخبار الأيام الأول 1",
+  "1ch_short": "1 Chr",
+  "2ch": "أخبار الأيام الثاني 2",
```
Pertempto, line 16: It looks strange, but this is the correct way to do numbers. See [here](https://www.wordproject.org/bibles/ar/index_en.htm#1).

packages/web/src/public/locales/ar/bible.json
Pertempto: Most of the name translations are computer generated. I went back and replaced a few from this [list](https://www.wordproject.org/bibles/ar/index_en.htm#1).

packages/web/src/shared/components/Dialog.tsx
```diff
@@ -31,7 +31,7 @@ const Dialog = forwardRef<DialogRef, DialogProps>(
       >
         {children}
         <button
-          className="absolute -right-1 -top-1 w-8 h-8 pt-2 pr-2"
+          className="absolute end-0 -top-1 w-8 h-8 pt-2 pe-1"
```
Pertempto, line 34: This is how I got around using negative values.

packages/web/src/main.tsx
```diff
@@ -1,23 +1,30 @@
...
 import { RouterProvider } from 'react-router-dom';
-import router from './app/router';
 import './app/i18n';
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import router from './app/router';
 import { FlashProvider } from './shared/hooks/flash';
 
+function App() {
+  const { i18n } = useTranslation();
+  document.body.dir = i18n.dir();
```
arrocke: We should only update this when the language changes. We can use `useEffect` to set up event listeners to [i18next events](https://www.i18next.com/overview/api#events)

packages/web/src/public/locales/ar/bible.json
arrocke: We can remove the short versions for now. In situations where we would use `t('gen', { context: 'short' })` it will fallback to the normal key if the `_short` key is missing.

packages/web/src/public/locales/ar/translation.json
arrocke, line 1: These locale json files have been updated in the past few PRs. Can you make sure the arabic ones are up to date?

packages/web/src/features/translation/VerseSelector.tsx
```diff
@@ -34,22 +34,26 @@ export function VerseSelector({ verseId, onVerseChange }: VerseSelectorProps) {
...
-        <Icon icon="arrow-left" />
-        <span className="sr-only">{t('previous_verse')}</span>
-      </button>
-      <button onClick={() => onVerseChange(incrementVerseId(verseId))}>
-        <Icon icon="arrow-right" />
-        <span className="sr-only">{t('next_verse')}</span>
-      </button>
+      {/* Use flex-row-reverse on rtl, so that the previous button is always
+          to the left of the next button. */}
+      <div className="flex ltr:flex-row rtl:flex-row-reverse gap-4">
```
arrocke: In a RTL language, we want the next button to be on the right. It's confusing for us, but natural in those languages. We should be able to remove this secondary flex box and leave it in the natural rtl order, but flip the icons.

Might be easiest to render two icons and show/hide the correct one with tailwind rather than trying to manage that with react

package.json
arrocke: I'm in total agreement about the ease of these new utilities. However, given that we've already run into a limitation of this unmaintained library, what do you think about using the code from the library directly in our tailwind config? Since we are going to be heavily using direction utilities, maybe we should have control over them. [Here](https://github.com/20lives/tailwindcss-rtl/blob/master/src/marginUtilities.js) is all that it would take to add margin utilities, including negative margins. That just wasn't released in this library. We wouldn't need to add everything the library supports to start, just the ones that we need.

arrocke (changes requested):
Nice work on figuring this out. It's working really well

arrocke:
I noticed this when preparing my update video: When in a RTL language, we still want the Hebrew translation line to be RTL. Right now, they are flipped back to LTR

packages/web/src/main.tsx
Pertempto: So I should create a local "dir" variable, update it in the event callback, and watch that variable in the useEffect? (I'm still trying to get comfortable with the react way of doing things)

packages/web/src/features/translation/VerseSelector.tsx
Pertempto: I wondered if this was the case or not

Pertempto: It just looked so strange to me 😆

Pertempto:
> I noticed this when preparing my update video: When in a RTL language, we still want the Hebrew translation line to be RTL. Right now, they are flipped back to LTR

Thanks for catching that! I remember noticing this, but then I forgot to address it.

packages/web/src/main.tsx
arrocke: Something like this should work:
```ts
useEffect(() => {
  function handler() {
    document.body.dir = i18next.dir()
  }
  i18next.on('initialized', handler)
  i18next.on('languageChanged', handler)
}, [i18next.dir, i18next.on])
```

That will attach event listeners to i18next and only apply changes when the language changes. Both the `dir` and `on` functions should be stable (meaning their reference doesn't change), so this effect will fire only once on first render and attach the listeners.

packages/web/src/public/locales/ar/bible.json
Pertempto: Done

packages/web/src/main.tsx
Pertempto: I tried this, but the handler never got called.

Pertempto: The problem is that both events trigger before the callback of `useEffect` is called.

Pertempto: I think my newest commit fixes it. Instead of using the `initialized` event, I just call the handler directly from the effect.

packages/web/src/public/locales/ar/translation.json
Pertempto: Done

packages/web/src/features/translation/VerseSelector.tsx
Pertempto: Done

package.json
Pertempto: I didn't even think of this. Good idea.

Pertempto: I think I brought over all the utilities we are using.

Pertempto:
I think I addressed everything you brought up. I'll be offline until Wednesday.

packages/web/src/tailwind-utilities/index.js
arrocke, line 1: What do you think about putting all of these utilities in the tailwind config directly rather than separating into a bunch of files? I think it will be easier to manage, but I'm curious what you think

arrocke (approved):

packages/web/src/tailwind-utilities/index.js
Pertempto: I don't like the idea of them being directly in `tailwind.config.js`, because we should keep that file simple to only config and external plugins we want to use. I will consolidate the different files into `tailwind-utilities.js`

PR #102, opened 2023-06-02
feat: restrict views based on user role

arrocke:
This should have been part of #96, but I forgot about it. This was based off my work in #99, but I decided to split it off since it may have separate conversation


packages/web/src/features/users/UsersView.tsx
```diff
@@ -20,7 +20,7 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import SelectInput from '../../shared/components/form/SelectInput';
 
 export default function UsersView() {
-  const session = useSession();
+  const session = useAuth({ systemRoles: [SystemRole.Admin] });
```
arrocke: I wanted a way to restrict who can get to certain views based on their roles. This used to be `useSession` to get the current user, but I coopted it let you specify which roles can access, and navigate away if the current user doesn't pass. What do you think of this approach?

Pertempto:
The target branch of this PR is `73-user-search`. Was that the intention?

packages/web/src/features/users/UsersView.tsx
Pertempto: So `useAuth` acts like a gatekeeper? I like it.

packages/web/src/shared/hooks/useAuth.ts
```diff
@@ -0,0 +1,67 @@
+import { useQuery } from '@tanstack/react-query';
+import { GetSessionResponse, SystemRole } from '@translation/api-types';
+import { useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
+import apiClient from '../apiClient';
+
+export interface UseAuthOptions {
+  authenticated?: boolean;
```
Pertempto: I don't see this field being used anywhere? It's usage confuses me.

packages/web/src/shared/hooks/useAuth.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    if (status === 'success') {
+      const { user } = data;
+      if (user) {
+        if (
+          authenticated === false ||
+          (systemRoles &&
+            !systemRoles.some((role) => user.systemRoles.includes(role)))
+        ) {
```
Pertempto: I'm not getting the `authenticated` usage here and line 43.

Pertempto: Ahhh, I might be understanding now. Need better explanation comments.

packages/web/src/shared/hooks/useAuth.ts
```diff
@@ -0,0 +1,67 @@
...
+  const { data, status } = useQuery(['session'], async () =>
+    apiClient.getSession()
+  );
+
+  useEffect(() => {
+    if (status === 'success') {
+      const { user } = data;
+      if (user) {
+        if (
+          authenticated === false ||
```
Pertempto: Why would an authenticated user not be allowed to see a certain page that authenticated users could see?

packages/web/src/shared/hooks/useAuth.ts
Pertempto: I added explanation comments to the best of my understanding. Please improve them if they are incorrect. I need additional clarification about the `authenticated === false` on line 35.

Pertempto (changes requested):
See my above comments.

packages/web/src/shared/hooks/useAuth.ts
arrocke: I clarified in the comments, but the basic use case is for login/signup pages, or for unauthenticated versions of pages on different routes.

packages/web/src/shared/hooks/useAuth.ts
arrocke: I rewrote the hook so that it's easier to understand both the options and the flow of logic. I didn't like that the name of the `authenticated` variable didn't reveal it's purpose without a comment. I also added a jsdoc comment to the hook itself that explains how to use it. Let me know if something is still missing.

arrocke:
> The target branch of this PR is 73-user-search. Was that the intention?

Yep. I had started this work in #99, but wanted to pull it out to be reviewed separately. When that PR merges, GitHub will automatically switch the target to `main`.

Pertempto:
> > The target branch of this PR is 73-user-search. Was that the intention?
> 
> Yep. I had started this work in #99, but wanted to pull it out to be reviewed separately. When that PR merges, GitHub will automatically switch the target to `main`.

Ok, so https://github.com/arrocke/gloss-translation/pull/99 needs merged first. I didn't get that.

packages/web/src/shared/hooks/useAuth.ts
```diff
@@ -4,10 +4,16 @@ import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
 import apiClient from '../apiClient';
 
-export interface UseAuthOptions {
-  authenticated?: boolean;
-  systemRoles?: SystemRole[];
-}
+export type UseAuthOptions =
```
Pertempto, line 7: This is a much more useful type definition! 👍🏽

Pertempto:
We should hide the "manage languages" dropdown item for non-admin.

Pertempto (changes requested):
We should hide the "manage languages" dropdown item for non-admin. Otherwise, this is looking great!

Pertempto (approved):

PR #107, opened 2023-06-15
feat: move language selector button

### LINKED ISSUE #95: More accessible place for interface language
Right now the interface language switcher is in the profile dropdown, which is hard to find and non-existing if you aren't logged in. We need to move this to a more visible place.

I'm thinking we add a footer and put this there as well as maybe consider automatically showing the language selector when you land on the site for the first time

Pertempto:
## What has changed

Moved the language selector button from the header to a new footer.

## Connected Issues

closes #95

### Change Language
* Click the language button in the footer
The language selector dialog should open and function nominally.

### First Load
* Open the app in a new private/incognito browser window
The language dialog should open automatically.

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/app/Footer.tsx
+++ b/packages/web/src/app/Footer.tsx
@@ -1,4 +1,4 @@
-import { useRef } from 'react';
+import { useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { DialogRef } from '../shared/components/Dialog';
 import { Icon } from '../shared/components/Icon';
@@ -6,8 +6,17 @@ import LanguageDialog from './LanguageDialog';
 import interfaceLanguages from './languages.json';
 
 export default function Footer() {
+  const i18nextLanguage = localStorage.getItem('i18nextLng');
+
   const languageDialog = useRef<DialogRef>(null);
   const { t, i18n } = useTranslation();
+
+  useEffect(() => {
+    if (i18nextLanguage == null) {
+      // This never occurs.
+      languageDialog.current?.open();
+    }
+  }, [languageDialog]);
 
   return (
     <div className="p-2 flex flex-row z-10 justify-end">
--- a/packages/web/src/app/Footer.tsx
+++ b/packages/web/src/app/Footer.tsx
@@ -4,19 +4,23 @@ import { DialogRef } from '../shared/components/Dialog';
 import { Icon } from '../shared/components/Icon';
 import LanguageDialog from './LanguageDialog';
 import interfaceLanguages from './languages.json';
+import useLocalStorageState from '../shared/hooks/useLocalStorageState';
 
 export default function Footer() {
-  const i18nextLanguage = localStorage.getItem('i18nextLng');
+  const [isFirstLoad, setIsFirstLoad] = useLocalStorageState<boolean>(
+    'needs-language-dialog',
+    true
+  );
 
   const languageDialog = useRef<DialogRef>(null);
   const { t, i18n } = useTranslation();
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


packages/web/src/app/Footer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const i18nextLanguage = localStorage.getItem('i18nextLng');
+
+  const languageDialog = useRef<DialogRef>(null);
+  const { t, i18n } = useTranslation();
+
+  useEffect(() => {
+    if (i18nextLanguage == null) {
+      // This never occurs.
+      languageDialog.current?.open();
+    }
```
Pertempto: > i18next already used a local storage property. If not defined, it will make an educated guess. Before i18next loads we can check ourselves, and show the modal if the local storage prop isn't defined. Then the modal will default to i18next's guess

I'm not sure how to access local storage before i18next

Pertempto: Even from `main.tsx`, that field is always set in local storage before the component function is called.

packages/web/src/shared/hooks/useLocalStorageState.ts
```diff
@@ -14,7 +14,7 @@ export default function useLocalStorageState<
   });
 
   useEffect(() => {
-    if (state) {
+    if (state != null) {
```
Pertempto, line 17: Just checking plain `state` doesn't work for `false` values.

packages/web/src/app/Footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTranslation } from 'react-i18next';
+import { DialogRef } from '../shared/components/Dialog';
+import { Icon } from '../shared/components/Icon';
+import LanguageDialog from './LanguageDialog';
+import interfaceLanguages from './languages.json';
+import useLocalStorageState from '../shared/hooks/useLocalStorageState';
+
+export default function Footer() {
+  const [isFirstLoad, setIsFirstLoad] = useLocalStorageState<boolean>(
+    'needs-language-dialog',
```
Pertempto: I ended up using a new local storage key. See this: https://github.com/arrocke/gloss-translation/pull/107#discussion_r1231607331

packages/web/src/app/Footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    if (isFirstLoad) {
+      setIsFirstLoad(false);
+      languageDialog.current?.open();
+    }
+  }, [isFirstLoad, languageDialog]);
+
+  return (
+    <div className="p-2 flex flex-row z-10 justify-end">
```
arrocke: Let's use the footer element here.

packages/web/src/app/Footer.tsx
arrocke: If we add this line to `i18next` before we call `init` then we can check local storage first:
```ts
export const initialLanguageChosen = !!localStorage.getItem('i18nextLng');
```

Now we can import that into the footer to show the modal. React refs don't need to be added to the dependency list of an effect, so we can guarantee that this effect only runs on first render
```ts
useEffect(() => {
  if (!initialLanguageChosen) {
    languageDialog.current?.open();
  }
}, []);
```

Pertempto:
Ready for review again.  
Will we add some color to the app at some point? The language dialog doesn't have much contrast with the rest of the content.

arrocke (approved):
That's a good question. I think we can revisit that once we've got users starting to use it. We need to have a conversation with Andrew and Sheldon about branding. I'm hoping that at some point a UX designer is willing to give a few hours a month to help out with that.

PR #109, opened 2023-06-18
build: simpler prisma executor

arrocke:
## What has changed

The upgrade of nx I did recently broke prisma. This is because prisma commands are often interactive, but nx wasn't connecting stdin to the prisma process. I found a way to write an executor that can run any prisma command and account for the interactivity.

## QA Steps

Try running `nx prisma api migrate reset`

## Post-Deployment

Update the deployment command in vercel to use `nx prisma api deploy`. Vercel builds will be broken until I do this


docs/db.md
```diff
@@ -9,15 +9,15 @@ Our database is postgresql@14.7 with prisma to manage the schema and migrations.
 1. Install postgresql version 14.7
 1. Create database for the project
 1. Set the `DATABASE_URL` env var in `packages/api/.env.local` to the postgres connection string. You may need to include a username and password in the connection string.
-1. Run `nx prisma-reset api` to scaffold the database schema.
+1. Run `nx prisma api migrate reset` to scaffold the database schema.
 1. Run `nx seed api` to insert static verse data into the database.
 
 ## Seed and Reset Data
 
 In order to reset the database to the current schema and clear all data, run the command:
```
Pertempto: Maybe we should include a comment about how this command takes several minutes?

Pertempto (approved):
It seemed to work for me.

PR #113, opened 2023-06-24
feat: language users

arrocke:
## What has changed

* Add language users
  * Database has three roles - viewer, translator, and admin
  * API and UI consider a user a member of a language if they have the viewer role
  * Members can additionally have either the translator or admin role, or both.
* Added multiselect input component

There is still more work to do in other PRs:
* Edit language roles
* Remove users
* Give UI access to language admins

## Connected Issues

closes #68

## QA Steps

1. Log in as a platform admin
2. Navigate to languages list
3. Open the manage page of a language
4. Add a user and give them a role
5. Check that user is created and added to language

## Post-Deployment

- [x] Nothing required

PR #117, opened 2023-07-02
feat: verify emails and handle bounces and complaints

### LINKED ISSUE #110: email verification
Before we can get SES turned on in AWS, we need email verification. This would include the following:
* Don't allow login emails to be sent until initial invitation is accepted
* If an email bounces, prevent the app from sending emails to that address

### LINKED ISSUE #115: accept invite flow
We need a separate view for accepting an invite, where the user can complete their profile. This will allow platform and language admins to invite a user with just an email. Later we may have more initial questions for new users.

arrocke:
## What has changed

* Added an invitation key that can be used to complete sign up
* Email is considered verified upon completing the profile form.
* Added route to handle SNS messages from SES when an email bounces or has a complaint.
* Users who have bounced, complained, or not accepted the invite cannot receive email.

## Connected Issues

closes #110, #115

## QA Steps

1. Log in to AWS
2. Navigate to https://us-east-1.console.aws.amazon.com/ses/home?region=us-east-1#/verified-identities/globalbibletools.com/send-test-email
3. Send bounced or complaint email
4. Check email status in users view as an admin

## Post-Deployment

- Set up SNS subscriber in production


packages/api/prisma/schema.prisma
```diff
@@ -14,22 +14,27 @@ datasource db {
...
+
 /// A user who can authenticate into the system.
 model AuthUser {
   /// The ID of the user.
-  id            String           @id @default(dbgenerated("generate_ulid()")) @db.Uuid
+  id           String           @id @default(dbgenerated("generate_ulid()")) @db.Uuid
   /// The display name of the user.
-  name          String?
-  /// The email address for logging in and notifications.
-  email         String?          @unique
```
arrocke: I removed the email from the user since it is now stored on the primary auth key.

packages/api/shared/Route.ts
```diff
@@ -69,26 +71,23 @@ export type RouteDefinition<Params, RequestBody, ResponseBody> = {
...
+   * Note that there are some schemas that typescript is unable to catch.
+   * The most common examples are:
+   * - Not adding optional properties to the schema.
+   * This won't return an error, but when those fields are provided, they won't be operated on.
+   * - Treating a property as required in the schema that is optional in the type.
+   * This will result in 422 errors if the field isn't provided.
+   * - Adding extra properties to the schema that aren't in the type.
+   * This could lead to unexpected 422 errors as well.
+   */
+  schema?: ZodType<RequestBody>;
```
arrocke, line 90: The fancy typescript types that made this conditionally required broke with a discriminated union as a request type. I had to make it optional, so if we forget to add a schema, the body won't be available in the request.

packages/api/shared/mailer.ts
```diff
@@ -1,16 +1,64 @@
...
-    // TODO: handle email verification
-    const response = await this.transporter.sendMail({
+  /**
+   * Send a transactional email to a user.
+   * @param email - The email to send along with the intended user.
+   * @param force - Send to unverified users. Should only be used to send invites.
+   * @throws `EmailNotVerifiedError` - If the user's email is not verified or has previously bounced or complained.
+   * @throws `MissingEmailAddressError` - If the user does not have an email address.
+   */
+  async sendEmail({ userId, subject, text }: EmailOptions, force = false) {
```
arrocke: mailer now takes a user id instead of an email. This let's it check whether the user's email is verified before we send the email

packages/web/src/features/users/AcceptInviteView.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const useInviteQuery = (token?: string) => {
+  const initialData = useLoaderData() as Awaited<
+    ReturnType<ReturnType<typeof acceptInviteLoader>>
+  >;
+  return useQuery({
+    ...createInviteQuery(token),
+    initialData,
+  });
+};
```
arrocke, line 45: This pattern of using both react-query and loaders from react-router I got from [this blog post](https://tkdodo.eu/blog/react-query-meets-react-router). I think we should use this as our primary data loading pattern. I'm going to document and streamline this in another PR.

Basically, the loader primes the cache and allows the view to wait to load until that request completes. Then after the view loads, the query will have data immediately because it is in the cache, but can be updated when it becomes stale.

Pertempto:
Migrations were added here and in #113, so I can't really test until that one is merged and the conflicts are resolved with this one.

arrocke:
@Pertempto Ok, this branch is ready to go now

Pertempto:
The testing steps aren't clear.  

> Send bounced or complaint email

To who? To one of the user emails in my local DB? If I do that, I get an unverified email error.

Pertempto:
Ok, I tried it in the preview and it worked. Am I supposed to do any testing in my local environment? 

When I tried inviting a user in my local environment, the link in the invite email just logged me in as the devyn account. I have to first log out of the devyn account, then try using the invite link. This seems like a bug.

Also, the invite link just shows up as plain text in the email. It is not formatted like a proper link like the emails from previous PRs.

Pertempto:
Is there any logic for re-sending an invite email?

If a user doesn't use the link in the original invite email, it seems that they are permanently locked out of their account. If they try to log in, the API gives an `EmailNotVerifiedError`

arrocke:
On the invite resending, we need them to accept the invite before we send them other emails otherwise we risk our sending reputation if we send to emails that bounce or complain about us. We can potentially resend the invite email as an admin action, but let's handle that separately

Regardless, I've got a couple things to look at

- [x] Email logs into wrong user
- [x] HTML emails
- [x] confirm that signin never triggers error

Pertempto:
Note regarding the "Email logs into wrong user" bug

When I tried it again, it made me think the bug has more to do with not properly logging out the previous user, than with any issue with the mail link. If I manually clicked the "Log out" button before using the invite link, it worked fine.

arrocke:
Yeah, that's what I noticed too. I'm ok with that behavior because the invite isn't being accepted, but the page that shows the invite is navigating away because a user is already logged in.

I pushed a commit that adds html emails back and replaces sending to ethereal.email with a console log if the EMAIL_SERVER env var isn't set. I needed that to develop offline and we don't need to test email delivery anymore.

I also am catching errors so attackers can't find out unverified users by spamming the login form.

arrocke:
@Pertempto This is ready for you to look at again

packages/api/pages/api/auth/login.ts
```diff
@@ -52,11 +52,17 @@ export default createRoute()
...
-        });
+        try {
+          await mailer.sendEmail({
+            userId: key.userId,
+            subject: 'GlobalBibleTools Login',
+            text: `Click the link below to log in to globalbibletools.com.\n\n${url.toString()}`,
+            html: `<a href="${url.toString()}">Log in<a> to globalbibletools.com`,
+          });
+        } catch (error) {
+          // We have to swallow errors here so that attackers don't know that a users email is unverified.
```
Pertempto:
```suggestion
          // We have to swallow errors here so that attackers don't know that a user's email is unverified.
```

Pertempto (changes requested):
The invite emails seem to be working well. I can easily invite users from the users page.

One problem, inviting users directly from a manage language page doesn't seem to work anymore. I get this error in the backend:
```
EmailNotVerifiedError: We can't send emails to a@a.cio because it is not verified.
    at Object.sendEmail (webpack-internal:///(api)/./shared/mailer.ts:54:116)
    at async Object.handler (webpack-internal:///(api)/./pages/api/languages/[code]/members/index.ts:123:13)
    at async eval (webpack-internal:///(api)/./shared/Route.ts:153:21)
    at async eval (webpack-internal:///(api)/./shared/Route.ts:212:21)
```

packages/api/shared/mailer.ts
```diff
@@ -1,16 +1,78 @@
...
+if (!process.env['EMAIL_SERVER'] && process.env.NODE_ENV === 'production') {
+  throw new Error('missing EMAIL_SERVER environment variable');
+}
+
+const transporter = process.env['EMAIL_SERVER']
+  ? createTransport({
+      url: process.env['EMAIL_SERVER'],
+    })
+  : {
+      async sendMail(options: SendMailOptions): Promise<void> {
```
Pertempto, line 14: I like this a lot. Opening ethereal.email just to login in my local env was becoming annoying.

packages/api/pages/api/languages/[code]/members/index.ts
```diff
@@ -102,13 +107,11 @@ export default createRoute<{ code: string }>()
...
-            email,
-          },
+          attributes: {},
         });
 
         const token = randomBytes(12).toString('hex');
-        await auth.createKey(user.userId, {
+        await auth.createKey(user.id, {
           type: 'single_use',
           providerId: 'email-verification',
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/members/index.ts]: If you invite a user from a language, they don't get prompted to enter their name. Maybe changing this to `providerId: 'invite-verification'` to  match `api/users/index.ts` would work?

It looks like there is a good amount of duplicate code between here and there. We could avoid some future inconsistency problems if we moved the duplicate code into a helper function.

Pertempto (changes requested):
Inviting from a language still doesn't work quite right.

packages/api/pages/api/languages/[code]/members/index.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/members/index.ts]: Ok, `providerId: 'invite-verification'` isn't the right fix. I tried it.

arrocke:
@Pertempto I think I got that fixed now. I pulled the invite code into a function since it's exactly the same. The original problem was that this PR was developed in parallel with the language users PR so I forgot to refactor that to the new invite pattern

Pertempto (approved):
The invite email works now.

At some point we'll need to update the UI to reflect user's permissions. For example, a language admin should be able to access the "manage languages" menu button, but instead it is only shown to users that have the admin system role.

Issue #84, opened 2023-05-12
handle diacritics gracefully


arrocke:
## acceptance criteria
* glosses with the same diacritics are treated as the same word
* autocomplete matches even when diacritics are in a different order

Pertempto:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize

Pertempto:
> glosses with the same diacritics are treated as the same word


What do you mean by this @arrocke? I'm not getting it.

Pertempto:
Maybe I should wait to do this one until after #106

arrocke:
> glosses with the same diacritics are treated as the same word

A letter can have multiple diacritics (ie in Hebrew שָּׁ is formed from the character \u05e9 plus these three diacritics in any order \u05c1, \u05bc, \u05b8). This means there are 6 ways of encoding that letter with those diacritics, but they render the same to the user. When a user types in a gloss, we need to connect to an existing matching gloss even if the diacritics are in a different order. Otherwise the number of ways a word has been glossed will be artificially inflated with what appears to be duplicates.

Pertempto:
Ok great! It appears that the built-in string normalization function handles that well. It'll just be a matter of figuring out how to modify the auto complete to support this. I'm going to put this back to "Ready" until I complete #106.

Pertempto:
Handled in #119

arrocke:
I think there may be a few additional things to this that can either be addressed in #119 or another:
1. We'll want to normalize on the backend as well under the assumption of not assuming the client has properly processed the input
2. We'll want to use 'NFD' normalization since that moves the diacritics to separate bytes in the string. That allows for searches that ignore the diacritics.
3. In the autocomplete, we want to match strings on just the characters, ignoring diacritics. For example if you were working with Hebrew, נֹתֵן (giving), and נָתַו (he gives) should both be matched by נתן without any of the vowel points.

Pertempto:
Ok I'll do it in a separate pr

PR #119, opened 2023-07-04
refactor: replace downshift with headless UI

### LINKED ISSUE #106: Replace downshift with headless-ui
The input width has some kind of glitch occasionally. 

image

Pertempto:
## What has changed

* Replaced `downshift` usage with `@headlessui/react`
* Replaced `AutocompleteInput` with `Autocomplete`

## Connected Issues

Closes #106

## QA Steps

1. Sign in with a translator account
2. Click the down icon for any word
    * A list of options should appear
3. Select an option
    * The new gloss should be saved
4. Type in the autocomplete input
    * The list of options should appear again, filtered by the input value.
5. Type a brand new gloss, then press tab
    * The new gloss should be saved

## Post-Deployment

- [ ] Nothing required


Pertempto:
I included the commits from `68-language-users`, so this is targeting that until #113 is merged.  
I didn't really write any new code. I was able to bring over pieces from the multi select and the old autocomplete components, and that basically covered everything.

Pertempto:
I forgot. This also include normalization code, so it should also cover #84. Test with unicode sequences you gave

> A letter can have multiple diacritics (ie in Hebrew שָּׁ is formed from the character \u05e9 plus these three diacritics in any order \u05c1, \u05bc, \u05b8). This means there are 6 ways of encoding that letter with those diacritics, but they render the same to the user. When a user types in a gloss, we need to connect to an existing matching gloss even if the diacritics are in a different order. Otherwise the number of ways a word has been glossed will be artificially inflated with what appears to be duplicates.

package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "@casl/ability": "6.5.0",
     "@casl/prisma": "1.4.0",
+    "@headlessui/tailwindcss": "^0.1.3",
```
arrocke: Let's lock this version down

packages/web/src/shared/components/form/Autocomplete.tsx
```diff
@@ -0,0 +1,123 @@
...
+  onBlur?(): void;
+  onChange?(value: string): void;
+  onCreate?(text?: string): void;
+  items: AutocompleteItem[];
+  defaultValue?: string[];
+  name: string;
+  hasErrors?: boolean;
+}
+
+const Autocomplete = ({
```
arrocke: Can we call this AutocompleteInput to be consistent with the other input components?

arrocke (approved):
Autocomplete looks good! I added a few comments to #84 with a few more ideas about diacritics. Up you to you if you want to include that here or in a separate PR.

package.json
Pertempto: Oops, I forgot

packages/web/src/shared/components/form/Autocomplete.tsx
Pertempto: Sure thing

PR #120, opened 2023-07-07
feat: handle diacritics gracefully

### LINKED ISSUE #84: handle diacritics gracefully

Pertempto:
## What has changed

Improved diacritics handling.

## Connected Issues

#84 
### Backend normalization
The easiest way to test this is to add these debug messages in `[wordId.ts]`:
```
      console.log('BEFORE:', new TextEncoder().encode(req.body.gloss ?? ""));
      const normalizedGloss = req.body.gloss?.normalize('NFD');
      console.log('AFTER:', new TextEncoder().encode(normalizedGloss ?? ""));
```
You'll also need to temporarily disable the frontend normalization in `AutocompleteInput.tsx`:
```
              setNormalizedInputValue(event.target.value)
```

1. Enter "è" as a gloss in the frontend. Select the create option in the dropdown.
2. Check the backend logs
  * You should see something like:
```
BEFORE: Uint8Array(2) [ 195, 168 ]
AFTER: Uint8Array(3) [ 101, 204, 128 ]
```
  * This means the "è" has been successfully decomposed.

### Ignore diacritics
1. Enter "tèst" as a gloss for a word. Make sure the gloss saved
1. Type "test" in the same autocomplete.
    * It should match "tèst", and have no create option
1. Repeat steps 1 and 2 on a different word, but initially save "test" as the gloss, and check that "tèst" matches it.

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/shared/components/form/AutocompleteInput.tsx
+++ b/packages/web/src/shared/components/form/AutocompleteInput.tsx
@@ -42,12 +42,12 @@ const AutocompleteInput = ({
     if (normalizedInputValue) {
       const filteredItems = items.filter((item) =>
         item.label
-          .normalize()
+          .normalize('NFD')
           .toLowerCase()
           .includes(normalizedInputValue.toLowerCase())
       );
       const noExactMatch = filteredItems.every(
-        (item) => item.label.normalize() !== normalizedInputValue
+        (item) => item.label.normalize('NFD') !== normalizedInputValue
       );
       if (noExactMatch && !!onCreate) {
         setFilteredItems([
@@ -88,7 +88,7 @@ const AutocompleteInput = ({
           <Combobox.Input
             {...props}
             onChange={(event) =>
-              setNormalizedInputValue(event.target.value.normalize())
+              setNormalizedInputValue(event.target.value.normalize('NFD'))
             }
             onBlur={onBlur}
             className="w-full py-2 px-3 h-10 rounded-b flex-grow focus:outline-none bg-transparent rounded"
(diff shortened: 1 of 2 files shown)


Pertempto:
I also tested with your example from https://github.com/arrocke/gloss-translation/issues/84#issuecomment-1620284889, and it seemed to work fine. Your "he gives" example has a different consonant the end, so it doesn't match.

packages/web/src/shared/components/form/AutocompleteInput.tsx
```diff
@@ -41,13 +41,14 @@ const AutocompleteInput = ({
...
-          .toLowerCase()
-          .includes(normalizedInputValue.toLowerCase())
+        ignoreDiacritics(item.label.normalize('NFD').toLowerCase()).includes(
+          ignoreDiacritics(normalizedInputValue.toLowerCase())
+        )
       );
       const noExactMatch = filteredItems.every(
-        (item) => item.label.normalize() !== normalizedInputValue
+        (item) =>
+          ignoreDiacritics(item.label.normalize('NFD')) !==
```
arrocke: Do you think the exact match should not ignore diacritics? I think this will make it difficult to add new glosses that have the same letters but different diacritics

arrocke (commented):
Just one question here

packages/web/src/shared/components/form/AutocompleteInput.tsx
Pertempto: Yeah, I agree. I wasn't clear about how exactly you wanted the autocomplete matching to work.

Pertempto: Try my newest commit.  

image

I think this behavior is good. It selects by default the other version, but still gives the user the option to create a new gloss if they want.

arrocke (approved):
I think this should work well. We'll find out when one of these languages starts using it

Issue #60, opened 2023-05-11
data import


arrocke:
## acceptance criteria
* As a language admin, I can select a language from the old app to import
* This data overwrites all of the existing glosses for the language

Pertempto:
What will this all involve? I know next to nothing about the data from the old app.

arrocke:
Here is how I think this could work:

1. Load language list by loading source from hebrewgreekbible.online. This file has a variable called `glossLanguageNames` that we can extract with a regex that we can use to render a dropdown to select the language to import
2. Data files are stored at URLs like https://hebrewgreekbible.online/EnglishGlosses/GenGloss.js. We can use the selected language and parse of all of the files with JSON.parse after removing `var glosses =` and comments.
3. Now we can insert glosses into our database from this data. Data is in a nested array of chapters, verses, words, word data. We can assume the order of words in a verse matches outs since our data source is the same.

Pertempto:
> Now we can insert glosses into our database from this data

What method should I use for inserting the data.

Pertempto:
Ahhh. At first I thought we were doing the data import in one sweep in a PR. Now I see that you want the data import to be triggered by the admins, and ran in the API. 👍🏽

Pertempto:
This is my first time writing API code for this project, so the first draft will be pretty terrible. Hopefully you can bare with me.

Pertempto:
@arrocke If it is ok with you, I'll do a series of smaller PRs for this one issue, each one building on the last. That way I can get feedback from you quicker, and get back on track earlier.

arrocke:
That's ok with me, just be clear in your PR what's included

Pertempto:
What library do you recommend using in node.js for making the requests to hebrewgreekbible.online

arrocke:
We are on node 18 so we can just use fetch!

Pertempto:
> We are on node 18 so we can just use fetch!

Oh this API is really nice and simple!

Pertempto:
@arrocke I was working on creating a separate language import page, but it just makes a lot of duplicate code.
When a user imports a language, they need to input the code and name (because the old names were all in English).


What do you think about this idea?
* We would have one language creation page
* At the top of the page, it gives a dropdown to select an import language
* The dropdown also has the option "None"
    * Maybe there is a better name
    * Selecting this option triggers the normal language creation API
* If they select an import language, we trigger the language import API

[...]

PR #121, opened 2023-07-07
feat: get import options

Pertempto:
## What has changed

Added the API route for getting language import options.
## Connected Issues

Relates to #60

## QA Steps

Send a `GET` request to `/api/import/languages` with the appropriate auth cookies.

## Post-Deployment

- [x] Nothing required


Pertempto:
@arrocke Here is my first attempt at a basic API route.

* Am I using things correctly?
* Am I naming things correctly?
* Am I organizing things correctly?
* How should this route handle the situation of an error accessing/parsing the language options from the external server?

packages/api/pages/api/languages/import/options.ts
```diff
@@ -0,0 +1,38 @@
...
+import { authorize } from '../../../../shared/access-control/authorize';
+
+const IMPORT_SERVER = 'https://hebrewgreekbible.online';
+
+export default createRoute()
+  .get<void, GetLanguageImportOptionsResponseBody>({
+    authorize: authorize({
+      action: 'read',
+      subject: 'AuthUser',
+    }),
```
arrocke: This route doesn't need authorization

arrocke (commented):
This route looks pretty good! How did you decide that this should be a route in our API? Another way to do this would be to load the list on the client side and then only provide a POST route to something like `/languages/{code}/import`

Pertempto:
>  How did you decide that this should be a route in our API?

I generally have the philosophy that anything involving data processing should be on the backend. This isn't the most intensive processing work, but in my rough testing it seems to take about half a second of time in my local server. If we put that half of second of processing time in the frontend code, the UX wouldn't be as good.

Pertempto:
If we did the request from the frontend, wouldn't we run into issues with cross-origin requests?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.preferences.importModuleSpecifier": "relative"
```
Pertempto: This seems to be your convention, correct? VS code kept suggesting full paths when it auto-added imports, and I didn't like having to manually change them all.

packages/api/pages/api/languages/import/options.ts
```diff
@@ -0,0 +1,33 @@
...
+            .split(',')
+            .map((name: string) => name.trim().replace(/['"]+/g, ''))
+            .filter((name: string) => name.length > 0);
+          // Return the options.
+          res.ok({ data: languageNames });
+          return;
+        }
+      } catch (error) {
+        console.error('Error retrieving options:', error);
+        // TODO: it would probably be better to throw an actual error?
```
Pertempto: How should this be handled? Is sending an empty list of options acceptable?

.vscode/settings.json
arrocke: It's a limitation of typescript with nx. All nx projects share the same typescript paths because that's how you reference other projects, this means you can't have aliases scoped to each project. So that means relative imports for internal dependencies and absolute for external dependencies. Perhaps this should be `non-relative` or `project-relative`.

The other question is whether we want to save vscode settings in the repo at all? I'm not sure what the right call is here. Open to your thoughts.

packages/api/pages/api/languages/import/options.ts
arrocke: If there is an error loading the data, we should definitely return one, but I think it should be a  500 error since it isn't an issue with the client's request, but on the "server" side. I think for now, we can leave it unhandled by removing the try catch, by default it should return 500. It's something only we would be able to fix, so if we ever see it on this route, we'll know how to track it down.

arrocke:
> If we put that half of second of processing time in the frontend code, the UX wouldn't be as good.

I'm not sure I agree with this. Regardless of where this takes place, we have to show a loading UI while the processing happens. Doing this in the API means an extra trip to the network, one to the API, and then another to the existing app, rather than just one to the existing app. Any time saved by doing it on the backend is lost in network latency.

> If we did the request from the frontend, wouldn't we run into issues with cross-origin requests?

With that being said, the cross-origin issue is a pain, so I think it best to do it like you have here.

arrocke (commented):
What do you think about moving this to `/api/import/languages`? My thinking is that the data from the old app belongs in a different api path than the languages in the existing app. Then to actually do the import that would be a post route on the language `/api/languages/{code}/import`

.vscode/settings.json
Pertempto: It's just a simple matter of discernment. For vs code settings that will help us write consistent code across developers, I don't see why there would be any problem. We're not going to add every vs code setting here.

Pertempto:
> What do you think about moving this to `/api/import/languages`? My thinking is that the data from the old app belongs in a different api path than the languages in the existing app. Then to actually do the import that would be a post route on the language `/api/languages/{code}/import`

Alright. I'd usually prefer to keep related routes under the same top-level route of the API.

I do get your point about 
>  data from the old app belongs in a different api path than the languages in the existing app

I'll move the API route.

Pertempto:
> I'm not sure I agree with this. Regardless of where this takes place, we have to show a loading UI while the processing happens.

But for simple practicality's sake, it's a lot easier to code the logic in the backend, right? Hypothetically, if there was no CORS consideration, if we did the processing in the frontend we'd want to use a web worker to avoid blocking the thread, and setting up a simple API route seems easier than setting up a web worker.

packages/api/pages/api/languages/import/options.ts
```diff
@@ -6,27 +6,22 @@ const IMPORT_SERVER = 'https://hebrewgreekbible.online';
...
-            .filter((name: string) => name.length > 0);
-          // Return the options.
-          res.ok({ data: languageNames });
-          return;
-        }
-      } catch (error) {
-        console.error('Error retrieving options:', error);
-        // TODO: it would probably be better to throw an actual error?
+      // Request the server root page.
+      const response = await fetch(IMPORT_SERVER);
```
arrocke: Should this route be removed?

arrocke (approved):
> But for simple practicality's sake, it's a lot easier to code the logic in the backend, right? Hypothetically, if there was no CORS consideration, if we did the processing in the frontend we'd want to use a web worker to avoid blocking the thread, and setting up a simple API route seems easier than setting up a web worker.

I think this is a fair point. For lower end devices, the cost of client side computation for long running processes can be much higher when compared to the network time.

Feel free to merge once this last comment is addressed

packages/api/pages/api/languages/import/options.ts
Pertempto: Not sure how I missed that 🤦🏽‍♂️

PR #123, opened 2023-07-14
feat: implement data import UI

Pertempto:
## Connected Issues

Related to #60

## QA Steps

1. Go to the manage page for a language
1. Click the import button
1. The backend will simulate 10 seconds of processing. The spinning indicator by the import button should spin while the processing occurs.
1. A success flash should be flashed

## Post-Deployment

- [x] Nothing required


packages/api-client/src/client.ts
```diff
@@ -1,14 +1,16 @@
...
 import Languages from './languages';
 import Users from './users';
 import Verses from './verses';
 import Words from './words';
 
-export { ErrorResponse };
+export * from './import';
 export * from './languages';
 export * from './verses';
 export * from './words';
```
Pertempto: What's the point of these

Pertempto:
Description coming later (maybe tomorrow)

Pertempto:
Todo - implement extra confirmation before allowing import

packages/api/pages/api/languages/[code]/import.ts
```diff
@@ -0,0 +1,28 @@
...
+    async handler(req, res) {
+      // TODO: implement import logic.
+      console.log('import language:', req.body.import);
+      // simulate the processing time
+      await new Promise<void>((resolve) => {
+        setTimeout(() => {
+          resolve();
+        }, 10000);
+      });
+      res.ok();
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/import.ts:25]: If you add a `throw Error();` above the `res.ok`, you can test that the error flash shows up correctly

Pertempto:
@arrocke This should handle most, if not all, the UI needed for #60.

packages/api/pages/api/languages/[code]/import.ts
```diff
@@ -0,0 +1,28 @@
...
+    })),
+    schema: z.object({
+      import: z.string(),
+    }),
+    async handler(req, res) {
+      // TODO: implement import logic.
+      console.log('import language:', req.body.import);
+      // simulate the processing time
+      await new Promise<void>((resolve) => {
+        setTimeout(() => {
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/import.ts:21]: This is obviously temporary. The next PR will handle the actual import process.

packages/api-client/src/client.ts
arrocke: The idea is that each of these would export the api types they are using, and then they'd get reexported here. That way you could import the types along side the client rather than just using the types. That doesn't turn out to be super useful in practice because the client gets instantiated once on the frontend, so there is no benefit in import types from `@pi/api-client` along side the client anymore. I'm fine if we want to remove them.

packages/api-client/src/import.ts
```diff
@@ -0,0 +1,14 @@
+import type { GetLanguageImportOptionsResponseBody } from '@translation/api-types';
+import type ApiClient from './client';
+
+export { GetLanguageImportOptionsResponseBody };
+
+export default class Import {
+  constructor(private readonly client: ApiClient) {}
+
+  languages(): Promise<GetLanguageImportOptionsResponseBody> {
```
arrocke: Can we call this `getLanguages`? I want to avoid confusion that `client.import.languages()` doesn't actually import anything

package.json
```diff
@@ -81,5 +81,8 @@
     "typescript": "5.0.4",
     "url-loader": "4.1.1",
     "zod": "3.21.4"
+  },
+  "prisma": {
+    "schema": "packages/api/prisma/schema.prisma"
```
arrocke, line 86: What's this for?

packages/api/public/spec/paths/languages/import.yml
```diff
@@ -0,0 +1,18 @@
+post:
+  tags:
+    - Language Import
+  summary: Import glossing data from hebrewgreekbible.online.
```
arrocke, line 4: This needs the language code parameter. You can use the same ref as the other routes that use the code parameter

packages/web/src/features/languages/ManageLanguageView.tsx
```diff
@@ -73,59 +73,62 @@ export default function ManageLanguageView() {
...
+                    {member.roles
+                      .map((role) => t('role', { context: role.toLowerCase() }))
+                      .join(', ')}
+                  </ListCell>
+                  <ListCell></ListCell>
+                </ListRow>
+              ))}
+            </ListBody>
+          </List>
+          <Link to="./import">{t('import_glosses')}</Link>
```
arrocke: We should add a prop to the Link component called `variant` that allows you to optionally specify `button` as the prop value. If that is true, then it renders the link like a button.

packages/web/src/features/languages/ManageLanguageView.tsx
```diff
@@ -73,59 +73,62 @@ export default function ManageLanguageView() {
...
-                  {member.roles
-                    .map((role) => t('role', { context: role.toLowerCase() }))
-                    .join(', ')}
-                </ListCell>
-                <ListCell></ListCell>
-              </ListRow>
-            ))}
-          </ListBody>
-        </List>
+        <div className="flex flex-col gap-8">
```
arrocke: What's the advantage of flex-col rather than adding margins between the Form, List, and Link?

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
```diff
@@ -0,0 +1,143 @@
...
+    useState<boolean>(false);
+
+  return (
+    <>
+      <View fitToScreen className="flex justify-center items-start">
+        <div className="mx-4 flex-shrink">
+          <ViewTitle className="flex">
+            <span>{language.data.name}</span>
+            <span className="mx-2">-</span>
+            <span>{language.data.code}</span>
```
arrocke: Could we make this something like `Import Glosses: {language.data.name}` to differentiate from the previous screen?

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
```diff
@@ -0,0 +1,143 @@
...
+      <ConfirmationDialog
+        title={t('import_glosses')}
+        description={t('import_glosses_confirmation_warning')}
+        confirmationValue={formContext.getValues()['import']}
+        isOpen={showConfirmationDialog}
+        setIsOpen={setShowConfirmationDialog}
+        onConfirm={async () => {
+          setIsSubmitting(true);
+          await onSubmit(formContext.getValues());
+          setIsSubmitting(false);
```
arrocke: Hmm. At my normal job, we have a way of opening a modal in the submit handler and then awaiting until it is closed before continuing with the submit. I really like how simple that pattern is. Something like:
```ts
async function onSubmit() {
  const { status } = await modal.show()
  if (status === 'ok') {
    // finish submitting
  }
}
```

There is already a SubmittingIndicator that we can use while the form is submitting and the submit button is disabled while submitting as well, so it would be nice if we could use this pattern.

If you like this pattern, we could accomplish this by using forwardRef on the dialog and providing a function `show` that returns a promise for when the modal is closed and returns whether the modal was canceled or confirmed. Then we'd have to put a ref on the component so that we can open it programmatically

arrocke (changes requested):
The modal looks really good! A few suggestions here

package.json
Pertempto: It tells Prisma studio where our Prisma schema is located

packages/web/src/features/languages/ManageLanguageView.tsx
Pertempto: I usually do it because I don't like individually specifying spacing on each item when I can just specify it once in the parent. Less code changes. If you want to update spacing you just have to change it at one place

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
Pertempto: Ok! I wanted to use submitting indicator but I couldn't since I'm not using normal form submission.

packages/web/src/features/languages/ManageLanguageView.tsx
Pertempto: > renders the link like a button

Do you mean that it should look just like our `Button` component?

packages/api/public/spec/paths/languages/import.yml
Pertempto: Done

packages/api-client/src/import.ts
Pertempto: Done

packages/api-client/src/client.ts
Pertempto: Done

packages/web/src/features/languages/ManageLanguageView.tsx
arrocke: Yes, I think that will look better

packages/web/src/shared/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,123 @@
...
+    () => ({
+      async open() {
+        return new Promise<boolean>((resolve) => {
+          setIsOpen(true);
+          // TODO: how do I call resolve when the confirm button is pressed, or the dialog is closed?
+
+          // setIsOpen(false);
+          // resolve(true);
+        });
+      },
```
Pertempto: @arrocke I got stuck on how to implement this.

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
Pertempto: Done

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
Pertempto: I think I know what you're asking for, but I got stuck in the implementation: https://github.com/arrocke/gloss-translation/pull/123#discussion_r1269212125

I pushed my commits so you can take a look.

packages/web/src/shared/components/ConfirmationDialog.tsx
arrocke: You can set resolve to a ref in this function. Then the click/submit handler for the modal can call the function from the ref

Pertempto: I got it working. I'd appreciate any advice on cleaning up the code.

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
Pertempto: Done

packages/web/src/features/languages/ManageLanguageView.tsx
Pertempto: Done. I tried to do in such a way that any changes to button styling in the future would be automatically brought across to `Link`

packages/web/src/public/locales/es/translation.json
```diff
@@ -22,5 +22,17 @@
...
+  "import_glosses_success": "¡Gloses importados con éxito!",
+  "import_glosses_title": "Importar Glosas: {{languageName}}",
+  "import_glosses_error": "Error de importación de glosas",
+  "import_description": "Seleccione un idioma para importar glosas de <1>hebrewgreekbible.online</1>.",
+  "import_glosses_confirmation_warning": "Esto sobrescribirá cualquier brillo asociado con el lenguaje. Puede llevar algún tiempo completar.",
+  "confirm_prompt": "Escriba '{{value}}' para confirmar su decisión.",
+  "confirm": "Confirmar",
+  "cancel": "Cancelar",
+  "language_name_required": "Ingrese el nombre del idioma.",
+  "language_code_required": "Ingrese el código de idioma."
```
Pertempto, line 37: Somehow these two terms got missed in the spanish file in a previous PR.

Pertempto:
Everything should be ready again @arrocke

packages/web/src/shared/components/actions/Link.tsx
```diff
@@ -1,10 +1,16 @@
...
 
-export function Link({ className, ...props }: LinkProps) {
-  return (
-    <BaseLink
-      className={`text-blue-600 focus:outline-none focus:underline hover:underline ${className}`}
-      {...props}
-    />
-  );
+type LinkProps = BaseLinkProps & {
+  variant?: 'base' | 'button';
```
arrocke: Can we use `default` instead of `base`?

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const { language, importLanguages } = useLoaderData() as {
+    language: GetLanguageResponseBody;
+    importLanguages: GetLanguageImportOptionsResponseBody;
+  };
+  const flash = useFlash();
+  const navigate = useNavigate();
+  const { t } = useTranslation();
+  const confirmationDialog = useRef<ConfirmationDialogRef>(null);
+  // This tracks if the backend is working on importing the glosses.
+  const [isImporting, setIsImporting] = useState<boolean>(false);
```
arrocke: We shouldn't need this import state anymore because the form will remain in loading state while the modal is open and the api request is in progress.

packages/web/src/shared/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+                  as="h3"
+                  className="text-lg font-medium leading-6 text-gray-900"
+                >
+                  {title}
+                </Dialog.Title>
+                <p className="text-sm text-red-500 font-bold">{description}</p>
+                <p id="prompt" className="text-sm text-gray-600">
+                  {t('confirm_prompt', { value: confirmationValue })}
+                </p>
+                <TextInput
```
arrocke: Can we wrap this in a Form, so we can get required validation and enter to submit? We'll also need to make the button a submit button

arrocke (changes requested):
Nice work with the modal ref! A few last things and then this should be good

packages/web/src/shared/components/actions/Link.tsx
Pertempto: Done

packages/web/src/features/languages/ImportLanguageGlossesView.tsx
Pertempto: I was using it to disable the `SelectInput` while the import was loading, but it looks like the user can't submit a new choice anyways so I'll remove it.

Pertempto: Done

packages/web/src/shared/components/ConfirmationDialog.tsx
Pertempto: Done

packages/web/src/shared/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,137 @@
...
+                      setEnableConfirmationButton(
+                        event.target.value === confirmationValue
+                      )
+                    }
+                    aria-describedby="prompt"
+                  />
+                  <div className="mt-4 gap-4 flex flex-row justify-end">
+                    <button
+                      className="focus:underline outline-none"
+                      type="button"
```
Pertempto, line 119: After I added the form, the cancel button started acting strange. I learned the hard way that [buttons default to `type="submit"`](https://stackoverflow.com/a/41904218)

arrocke (approved):

PR #125, opened 2023-07-15
refactor: role based access control for views and actions

### LINKED ISSUE #69: enforce language access
## acceptance criteria
* As members of a language, we should be the only ones besides system admins, who can see the data for our language
* I should only see the list of languages that I have access to

## notes
* We will tackle the readonly version for unauthenticated users later

arrocke:
## What has changed

This focuses on authorizing views and actions within views. I was working on making sure that language admins have access to everything they need, but this is a big enough change that I'll leave the rest for a separate PR.

* Introduces a way to authorize each route according to the users role in the platform and individual language.
* Introduces a way to authorize parts of a view like managing/creating a language
* This requires language memberships to be returned in the user session
* Since users may not be authorized to view a language, we need a proper home view that has no authorization restrictions

## Connected Issues

closes #69 

## QA Steps

1. Log in as an admin and verify that you have access to everything
2. Create a language admin and login. Verify you have access to that language, but not others
3. Create a language translator and login, Verify you have access to that language, but not manage it

## Post-Deployment

- [x] Nothing required


packages/web/src/shared/accessControl.tsx
```diff
@@ -0,0 +1,131 @@
...
+    roles: LanguageRole[];
+  }[];
+}
+
+export type Action = 'create' | 'read' | 'translate' | 'administer';
+export type SubjectType = 'User' | 'Language';
+export type Subject = SubjectType | { id: string };
+export type Policy = PureAbility<[Action, Subject]>;
+
+export function createPolicyFor(user?: Actor) {
```
arrocke, line 22: Just like the backend, we can create a policy for access control with casl.

packages/web/src/shared/accessControl.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { redirect } from 'react-router-dom';
+
+interface Actor {
+  id: string;
+  systemRoles: SystemRole[];
+  languages: {
+    code: string;
+    roles: LanguageRole[];
+  }[];
+}
```
arrocke, line 15: But now there is more information on the Actor. This data is loaded in the session

packages/web/src/shared/accessControl.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  return build();
+}
+
+export async function authorize(
+  action: Action,
+  subject: SubjectType | { type: SubjectType; id: string }
+) {
```
arrocke: This function can be called from a loader. It loads the session and throws and error if the user doesn't have access. That error will need to be handled to redirect back to the home page

packages/web/src/shared/accessControl.tsx
```diff
@@ -0,0 +1,131 @@
...
+export interface UserCanProps {
+  subject: SubjectType | { type: SubjectType; id: string };
+  action: Action;
+}
+
+export function UserCan({
+  children,
+  action,
+  subject,
+}: UserCanProps & { children: ReactNode }) {
```
arrocke: This component works similarly, but it will render its children if the user has access. This can be used for situations where some users can view a bunch of resources, but a subset can perform actions on them.

packages/web/src/features/languages/router.tsx
```diff
@@ -9,20 +10,30 @@ import NewLanguageView from './NewLanguageView';
...
     element: <NewLanguageView />,
   },
   {
     path: 'languages/:code',
-    loader: manageLanguageViewLoader,
+    loader: async ({ params }) => {
+      const code = params.code as string;
+      await authorize('administer', { type: 'Language', id: code });
+      return manageLanguageViewLoader(code);
+    },
```
arrocke: Here is an example of authorize in use with a data loader

packages/web/src/features/languages/LanguagesView.tsx
```diff
@@ -39,20 +38,25 @@ export default function LanguagesView() {
...
               <ListRow key={language.code}>
                 <ListCell header>{language.name}</ListCell>
                 <ListCell>
-                  <Link to={`./${language.code}`}>{t('manage')}</Link>
+                  <UserCan
+                    action="administer"
+                    subject={{ type: 'Language', id: language.code }}
+                  >
+                    <Link to={`./${language.code}`}>{t('manage')}</Link>
+                  </UserCan>
```
arrocke: And here is an example of the component in use

arrocke (commented):
@Pertempto Could you give this a preliminary review? This is a significant change to how access control can work on the frontend, but it's flexible enough to handle roles on languages, not just system roles. You don't need to do a code review yet just feedback on the idea, I've still got a lot of work to clean things up.

Pertempto:
> @Pertempto Could you give this a preliminary review? This is a significant change to how access control can work on the frontend, but it's flexible enough to handle roles on languages, not just system roles. You don't need to do a code review yet just feedback on the idea, I've still got a lot of work to clean things up.

It looks pretty good to me! The translator role isn't working quite right. I get 403 errors from the backend when I try to go to the translate page.

arrocke:
@Pertempto I think this is ready for a full review now

Pertempto:
I'm going to try to wrap up https://github.com/arrocke/gloss-translation/pull/123 before I review this.

packages/web/src/app/NotFound.tsx
```diff
@@ -1,14 +1,20 @@
...
\ No newline at end of file
+  const message =
+    error instanceof NotFoundError
+      ? 'Not Found'
+      : error.statusText || error.message;
+
+  return (
+    <div className="absolute w-full h-full flex flex-col items-center justify-center">
+      <h1 className="font-bold text-xl mb-6">Woops!</h1>
+      <p className="mb-2">An unknown error has occurred.</p>
```
Pertempto, line 16: Do we want to translate these?

packages/web/src/app/NotFound.tsx
```diff
@@ -1,14 +1,20 @@
...
+      <p className="italic">{message}</p>
```
Pertempto, line 17: It would be good to include a link to the home page.

packages/web/src/app/NotFound.tsx
Pertempto, line 16: Do we want to translate these?

Pertempto (changes requested):
# Bugs
- A system admin still sees a writable language name and "update" button, even when they don't have the language admin role to update the language.
- The "translate" link shows up in the nav bar, even for a user that only has the "admin" role for that language
- I can update the name of a language with a user that has both "admin" and "translator" language roles, but NOT with a user that has only "admin"

packages/web/src/app/Header.tsx
```diff
@@ -50,11 +57,11 @@ export default function Header({ language, onLanguageChange }: HeaderProps) {
                 </DropdownMenuButton>
               ))}
             </DropdownMenuSubmenu>
-            {session.user?.systemRoles.includes(SystemRole.Admin) && (
+            <UserCan action="administer" subject="Language">
```
Pertempto: This item showed up for an user that ONLY had the translate role on one language. They had no admin role on any language.

packages/web/src/app/NotFound.tsx
arrocke: I'm inclined to clean up the not found page in a separate PR once we know what more of the errors are going to be. Are you ok if I open a separate issue for the error page?

arrocke:
> A system admin still sees a writable language name and "update" button, even when they don't have the language admin role to update the language.

I was thinking we would let admins have access to everything except translating

Everything else should be fixed

packages/web/src/app/NotFound.tsx
Pertempto: Sounds good 👍🏽

Pertempto:
> I was thinking we would let admins have access to everything except translating

OK, we should clearly document the access system sometime.

arrocke:
@Pertempto I merged in main which has your import changes if you want to review how the access control works with those views. I also added to the auth docs about how our access control works. Let me know if I missed something there.

Pertempto:
> @Pertempto I merged in main which has your import changes if you want to review how the access control works with those views. I also added to the auth docs about how our access control works. Let me know if I missed something there.

Great! I hope to review this evening.

packages/web/src/app/NotFound.tsx
Pertempto: I don't see the new issue?

Pertempto:
All the bugs I seen before seem to be fixed now 👍🏽

Pertempto:
I see one bug that is kind of unrelated to this PR, but we should fix sometime - updating the name of a language doesn't update the name shown on the top of the page until you refresh your browser.

docs/auth.md
```diff
@@ -1,5 +1,39 @@
...
+### Subjects
+
+While both the api server and the web client have `Language` and `User/AuthUser` subjects, how these subjects are narrowed works differently. On the server, a prisma query can be used to narrow the access to actions on particular subjects. On the web client, the filter is limited to lists of subject IDs generated by session data.
+
+### Policy
+
+At present the policy is this:
+
+- Platform Admins - can do anything except translate a language
+- Language Admin - can view language translations, and manage language members
```
Pertempto: The admin role goes past managing users for a language, they can also update it's metadata (currently just the name), and import glosses. Maybe we should include some kind of wording the encompasses those actions? 

```suggestion
- Language Admin - can view language translations, manage language information/members, and import glosses from the previous site
```

Pertempto (changes requested):
I managed to find a bug

1. Give a user translator access to a language other than `en` English
2. Login as the user
3. Try to go to the translate page
The console gives lots of 403 errors about accessing `/api/languages/en/verses/01001001/glosses`, because the user doesn't have access to the `en` language.

I believe it is caused by line 21 of `TranslationView.tsx`
```
 () => apiClient.verses.findVerseGlosses(params.verseId, 'en')
```

Shouldn't every user have access to just READ the `en` glosses?

arrocke:
This is a good catch. I think for now, the best approach is to allow all languages to be read. At some point we may build something to publish languages, so that the language members can decide which verses have their glosses public.

So a couple changes in that last commit:
* replace `<UserCan />` with a hook `useAccessControl`. A function can be used to both conditionally render components, but also to pass booleans to components
* Remove access control for reading language glosses and the list of languages
* Remove access control for the translation view, but removing the autocomplete if user can't translate

Pertempto:
> I think for now, the best approach is to allow all languages to be read.

I agree!

Pertempto (approved):
I like those most recent changes a lot! It seems much more intuitive for all users to have read-only access to all languages.

Issue #124, opened 2023-07-15
Switch to postmark for sending emails


arrocke:
I'm not able to get SES production access without a bunch more work. They won't even tell me what I need to do. I propose we switch to postmark for email delivery. They are a transactional email service that I've used in the past that I like. 100 emails a month are free, and after that its $15 for 10,000 emails a month. I think if we switch to passwords for authentication we could stay in that free tier for a while

arrocke:
@Pertempto What do you think about this?

Pertempto:
My only reservation is that we invested a lot of work into the email link login system so it doesn't sound great to scrap all that. How much rework will be involved in switching to a password system?

It would also be interesting to look at using third party auth tools like sign in with Google. That definitely shouldn't be our primary system but it could be useful to our users in that they don't have to create another password.

arrocke:
With lucia, it won't be too bad. We just need to ask for a password on invite accept, and check it with the login screen and route. At this point the investment is a sunk cost so here is how I'm thinking about it.
1. We can move to postmark for email sending. This is pretty simple: set up postmark and point our sever to their SMTP server rather than AWS. This much is a no brainer at this point, since getting AWS to let us send emails is pretty difficult
2. The reasons we would switch to passwords are separate. Either to save money or because we think the user experience is better. Frankly after using the magic links for a while, I don't prefer them because my password manager has made logging in with a password a breeze.

Pertempto:
Sounds good to me! I'm fine with using passwords for login

arrocke:
Lol, after I sent a frustrated response in the support ticket about their process, they gave us access. I'm going to close this

Pertempto:
> Lol, after I sent a frustrated response in the support ticket about their process, they gave us access. I'm going to close this

Hahaha, nice!

PR #126, opened 2023-07-16
refactor: use passwords for auth

arrocke:
## What has changed

* Added passwords to accept invites and login routes and views
* Added length and confirmation validators to the text input
* Modified the error message component to be more flexible. When you use it, you have to specify each error message.
* Pulled out the users translation strings into a separate json file

## Connected Issues

relates to #124 

## QA Steps

1. Add the following value to `hashed_password` in your admin user in the database:
`s2:h2I35PXwsp81qZNX:fcb5e8e8bbb2b6bd93cad7d4af84c8057f58254d45574a3809fb3455ca88a7ee6d4690c89ea6c63e66bda1f5229da8b95f4227e5a6e21cac6acd83dcd4b730c7`
2. Try logging in with password `asdf1234`
3. Invite a user and confirm that the password you used to sign up works to log in

## Post-Deployment

We'll need to add passwords to existing users


packages/api/public/spec/paths/auth.yml
```diff
@@ -1,18 +1,4 @@
...
-        in: query
-        description: the single use token
-      - name: redirectUrl
-        in: query
-        description: where to redirect to when login succeeds.
-    responses:
-      302:
-        description: Redirect
   post:
     summary: Login with magic link
```
Pertempto: This needs updated

packages/web/src/public/locales/ar/translation.json
```diff
@@ -1,47 +1,28 @@
 {
-  "language_one": "لغة",
+  "language_one": "اللغة",
```
Pertempto, line 2: Why so many changes in this file?

Pertempto (changes requested):
I'm impressed with how few API changes were required to support passwords!  

# Improvements
* I think we could have a better login error message when users give an incorrect password.
* The browser doesn't recognize the email field on the accept invite page, because it isn't an input field. This breaks the built-in password manager because it tries to save the username as the email.  
    * I think maybe we could fix this by making the email a read-only input field? Maybe some additional annotations are required to make the browser recognize it.
    * I was testing this in Vivaldi, a chromium browser, so this might be specific to the chromium family tree of browsers.
    * EDIT: the same behavior occurs in Firefox.

# Bugs
* When a user accepts an invite and submits their username/password, we still get the incorrect flash.
- If a user has no password in the DB, they can login with an empty password. Even a brand-new invited user can do this before they use their invite link.

packages/web/src/public/locales/ar/translation.json
arrocke: I didn't want to reorganize the file, so I just asked chatgpt to translate the entire file. It apparently came up with something different than google translate for a few of the keys

arrocke:
@Pertempto Great feedback. I think I've addressed it all

packages/web/src/public/locales/ar/translation.json
Pertempto: Ahh, interesting! I find it tedious to use Google Translate manually, so I usually generate translations with this extension: https://marketplace.visualstudio.com/items?itemName=funkyremi.vscode-google-translate

Pertempto (approved):
This is looking pretty good!

One small thing that needs finished before the merge:
- `invite_user` doesn't get translated correctly in `ManageLanguageView`

PR #127, opened 2023-07-28
feat: import gloss data from old server

### LINKED ISSUE #60: data import

Pertempto:
## What has changed

Actually implemented the logic for importing glosses from https://hebrewgreekbible.online

## Connected Issues

closes #60 

## QA Steps

1. Create a language or use an existing one
2. Go to the import page for your language
3. Select a language from the old site in the dropdown
4. Click "import glosses" and confirm the dialog
5. Wait for the import process to complete. It will probably take a few notes
6. Go to Genesis 1:1 or some other verse. Check that glosses were imported correctly
    - You may have to search for a verse with glosses. Many of the languages from the old site are incomplete

## Post-Deployment

- [ ] Nothing required


Pertempto:
NOTE: I am nowhere close to finishing this. I just started. I opened this PR so I could ask you questions 😁

packages/api-types/src/index.ts
```diff
@@ -1,3 +1,5 @@
+export { bookKeys } from './book-keys';
```
Pertempto: Where should we put little bits of code/data that we need to share between the frontend and backend, to avoid duplicate code? Should we have a new nx package? I'm not sure how to set that up.

packages/web/src/features/translation/verse-utils.ts
```diff
@@ -130,7 +130,7 @@ export function incrementVerseId(verseId: string) {
  * @returns The translated name of the book.
  */
 export function bookName(bookId: number, t: TFunction) {
-  return t(bookKeys[bookId - 1], { ns: 'bible' });
+  return t(bookKeys[bookId - 1].toLowerCase(), { ns: 'bible' });
```
Pertempto, line 132: I decided it was simpler to manually capitalize the keys in `bookKeys` and then lowercase them here, rather than trying to capitalize them in the API logic. The capitalization logic gets a little more complicated when you're wanting to convert from `1ch` to `1Ch`.

This version of `bookKeys` is also more accurate, because it matches the filenames in `data/morphology`, AND the JS gloss filenames (for example: `https://hebrewgreekbible.online/EnglishGlosses/1ChGloss.js`)

Pertempto:
@arrocke Please take a look when you get the time. (a look at my questions/comments, not the code)

Pertempto:
Great news! I got the basic parsing logic working that converts the JS code from the old server to valid JSON.

The main thing left to do is actually insert the glosses.

packages/api/pages/api/languages/[code]/import.ts
```diff
@@ -14,15 +18,35 @@ export default createRoute()
...
+      }
       res.ok();
     },
   })
   .build();
+
+/**
+ * Parse JSON data from the JS gloss scripts.
+ */
+function parseGlossJs(jsCode: string) {
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/import.ts:105]: Over commented this function on purpose! It uses a lot of regular expressions, and regex code is usually NOT very intuitive to read.

Pertempto:
The parsing runs faster than I expected. We'll have to see how much overhead inserting thousands of glosses into the DB adds.

packages/web/src/features/translation/verse-utils.ts
arrocke: I think this is a good approach. It might be good to match the keys in bible.json to the same format so they are all the same.

packages/api-types/src/index.ts
arrocke: Can we put this in the data folder next to the morphology script? Then the morphology script can use it as well.

packages/web/src/features/translation/verse-utils.ts
Pertempto: I thought about that, but our current convention is to have all lowercase translation keys. I'd rather keep it that way.

Pertempto:
It turns out our original assumption about checking for length 6 in the morphology data wasn't accurate. See Genesis 1:20.

```
["וַיֹּ֣אמֶר","way·yō·mer","And said","Conj-w | V-Qal-ConsecImperf-3ms","559 [e]",""],
["אֱלֹהִ֗ים","’ĕ·lō·hîm,","God","N-mp","430 [e]"," ､"],
["תַּֽדְשֵׁ֤א","taḏ·šê","let bring forth","V-Hifil-Imperf.Jus-3fs","1876 [e]",""],
["הָאָ֙רֶץ֙","hā·’ā·reṣ","the earth","Art | N-fs","776 [e]",""],
["דֶּ֔שֶׁא","de·še,","grass","N-ms","1877 [e]"," ､"],
["עֵ֚שֶׂב","‘ê·śeḇ","the herb","N-ms","6212 [e]",""],
["מַזְרִ֣יעַ","maz·rî·a‘","[that] yields","V-Hifil-Prtcpl-ms","2232 [e]",""],
["זֶ֔רַע","ze·ra‘,","seed","N-ms","2233 [e]",""],
["עֵ֣ץ","‘êṣ","[and] the tree","N-msc","6086 [e]"," ､"],
["פְּרִ֞י","pə·rî","fruit","N-ms","6529 [e]",""],
["עֹ֤שֶׂה","‘ō·śeh","[that] yields","V-Qal-Prtcpl-ms","6213 [e]",""],
["פְּרִי֙","pə·rî","fruit","N-ms","6529 [e]",""],
["לְמִינ֔וֹ","lə·mî·nōw,","according to its kind","Prep-l | N-msc | 3ms","4327 [e]"," ､"],
["אֲשֶׁ֥ר","’ă·šer","whose","Pro-r","834 [e]",""],
["זַרְעוֹ־","zar·‘ōw-","seed [is]","N-msc | 3ms","2233 [e]",""],
["ב֖וֹ","ḇōw","in itself"],
["עַל־","‘al-","on","Prep","5921 [e]",""],
["הָאָ֑רֶץ","hā·’ā·reṣ;","the earth","Art | N-fs","776 [e]","–"],
["וַֽיְהִי־","way·hî-","and it was","Conj-w | V-Qal-ConsecImperf-3ms","1961 [e]",""],
["כֵֽן׃","ḵên.","so","Adv","3651 [e]"," ."]
```

Pertempto:
`["ב֖וֹ","ḇōw","in itself"],`


Only has 3 items, but it is included in the gloss data https://hebrewgreekbible.online/EnglishGlosses/GenGloss.js


Should I just ignore items like this?

Pertempto:
Where did you get your morphology data? Was it from hebrewgreek.online? So far it doesn't seem to match up very well.

Pertempto:
My errors were caused by my logic for ignoring the `.length != 6` stuff was wrong. I seem to be having success importing data now!! 🎉 

I'll clean up the code tomorrow hopefully.

packages/api/pages/api/languages/[code]/import.ts
```diff
@@ -14,15 +19,103 @@ export default createRoute()
...
+      //       ton of requests. use all keys when things are working
+      for (const key of bookKeys) {
+        console.log(key);
+        const importUrl = `${importServer}/${req.body.import}Glosses/${key}Gloss.js`;
+        const response = await fetch(importUrl);
+        const jsCode = await response.text();
+        const bookData = parseGlossJs(jsCode);
+        const bookId = bookKeys.indexOf(key) + 1;
+        const referenceData =
+          require(`../../../../../../data/morphology/${key}`) as string[][][];
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/import.ts]: I had to temporarily disable the pre-commit hook to ignore the linter error about using `require`. How did you make it pass in `data/morphology.ts`?

Comment (arrocke) [reply packages/api/pages/api/languages/[code]/import.ts]: A dynamic import like that isn't going to work with a bundler. Instead we should update `data/morphology.ts` to import all of the json files and export as an object that maps the book key to the data. That way the bundler can include all of the data files in the build

Pertempto:
It took 398 seconds on my machine to import the French glosses, but it doesn't have glosses for the whole Bible.

packages/api/pages/api/languages/[code]/import.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/import.ts]: Ok!

packages/api-types/src/index.ts
Pertempto: I need to remember to do this tomorrow morning.

Pertempto: Done

packages/api/pages/api/languages/[code]/import.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/import.ts]: Done

Pertempto:
It took about 10 minutes to import the Arabic glosses, which seems to be only OT

Pertempto:
I updated from many `create` calls to one `createMany` call, and Arabic now takes 37 seconds 😁 

Seeing that the NT is shorter than the OT, I think we can be pretty confident that a normal import process would usually take less than 1 minute.

Pertempto:
This is ready @arrocke! I'm going to reset my DB and check that the seed script still works.

Pertempto:
Ok the seed script works, after a few adjustments to the import statements.

Pertempto:
@arrocke Can you remind me of the SQL i need to run to insert a default user into the data? I don't want to use the dump file. Maybe we could add some logic to `seed.ts` that inserts a default user.

arrocke (changes requested):
Good news: code looks good. nice work!

Bad news: the vercel build is too big when we include all of the morphology data in the bundle. One option is to make a second request to `https://hebrewgreekbible.online/files/<book-code>.js` to load the morphology data into memory. It will slow down the process, but it should be at most twice as long. A 2 minute import done once for each language does not seem to bad to me.

arrocke:
Here's what you need to insert a user with email `admin@example.com` and password `asdf1234`
```sql
INSERT INTO "User" ("id", "name", "emailStatus")
VALUES ('34c65977-f311-430b-af2a-67cee04b6faa', 'Admin User', 'VERIFIED');

INSERT INTO "UserAuthentication" ("id", "hashed_password", "user_id", "primary_key")
VALUES (
	'username:admin@example.com',
	's2:h2I35PXwsp81qZNX:fcb5e8e8bbb2b6bd93cad7d4af84c8057f58254d45574a3809fb3455ca88a7ee6d4690c89ea6c63e66bda1f5229da8b95f4227e5a6e21cac6acd83dcd4b730c7',
	'34c65977-f311-430b-af2a-67cee04b6faa',
	TRUE
);

INSERT INTO "UserSystemRole" ("userId", "role")
VALUES ('34c65977-f311-430b-af2a-67cee04b6faa', 'ADMIN');
```

Pertempto:
It is a little tricky because the reference files contain multiple `var <name-here>=` sections, and my parsing function doesn't know how to handle that. Trying to find the appropriate changes.

Pertempto:
@arrocke I believe this works now!

arrocke (approved):
Nice work!

PR #130, opened 2023-07-28
fix: seed script word ID

### LINKED ISSUE #129: Word ID is 1 off
The `wordId` column in the `Word` table is off by one. The word ID for the first word of Genesis 1:1 is `0100100102`, instead of the expected `0100100101`

Pertempto:
## Connected Issues

Closes #129 
## QA Steps

1. Run the seed script.
2. Ensure that the first entry in the word table has word ID `0100100101`, NOT `0100100102`.

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/api/prisma/migrations/20230728141110_fix_word_ids/migration.sql
+++ b/packages/api/prisma/migrations/20230728141110_fix_word_ids/migration.sql
@@ -1,0 +1,32 @@
+-- We have to remove these constraints so that we can mess with ids without violating them.
+ALTER TABLE IF EXISTS public."Gloss" DROP CONSTRAINT IF EXISTS "Gloss_pkey";
+ALTER TABLE IF EXISTS public."Gloss" DROP CONSTRAINT IF EXISTS "Gloss_wordId_fkey";
+ALTER TABLE IF EXISTS public."Word" DROP CONSTRAINT IF EXISTS "Word_pkey";
+-- For both words and glosses we have to extract the last two digits, decrement them, and then ensure they are padding with up to one 0
+UPDATE "Word"
+SET "id" = CONCAT(
+    SUBSTRING("id", 1, 8),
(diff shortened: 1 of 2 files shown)


Pertempto:
@arrocke I haven't tested this, but I think this should fix #129 

Could you generate the new seed dump file?

Pertempto:
Line 53:

```
const filteredWords = words.filter((word) => word.length === 6);
```

Is going to cause us problems, because the gloss files include these extra words. This makes the word IDs out of sync. For example see https://hebrewgreekbible.online/EnglishGlosses/GenGloss.js

Genesis 1:5 has 14 entries there, but we only import 13 words. In this case the extra gloss is at the end, but what if the extra gloss was in the middle? This throws word IDs way off.


I wonder if I could work around it by importing all the morphology data in my import logic, and checking for the length 6 thing in our morphology data before importing a gloss.

arrocke:
Nice catch! I think I like the approach of using the morphology data. That's already in the data folder, so we shouldn't need a network request for that.

I'll generate a new seed file, and see if there is a way to migrate the data in place without having to overwrite the entire database

arrocke:
I wrote a custom prisma migration that should fix this in preview and production without have to overwrite the database and lose users. You can try it for yourself by running `nx prisma api migrate dev` locally.

arrocke (approved):

arrocke:
@Pertempto I approved the review, but you should look at the migration I added before merging this.

Pertempto:
> I wrote a custom prisma migration that should fix this in preview and production without have to overwrite the database and lose users. You can try it for yourself by running `nx prisma api migrate dev` locally.

Great! This is why I asked you to help 😁 I was hoping there was a way to do this.

Pertempto:
It worked for me! Migration looks good.

PR #132, opened 2023-07-28
feat: modify and remove language members

### LINKED ISSUE #70: language admins
## acceptance criteria
* Users on a language can have one or both of the edit or admin permissions.
* Language admins, can add other users, and change the permissions of other users on the language

arrocke:
## What has changed

* multiselect for modifying a language members roles
* button to remove member from language
* added secondary and tertiary variants for buttons

## Connected Issues

closes #70 

## QA Steps

1. add a member to a language as an admin
2. verify that the user can see the manage language view and can't translate
3. now add the translator role
4. verify that the user can translate
5. now remove the admin role
6. verify that the user can't access manage language view
7. now remove the member
8. verify that the user can only read the translation view and can't translate

## Post-Deployment

- [x] Nothing required


packages/web/src/shared/components/actions/Button.tsx
```diff
@@ -3,12 +3,27 @@ import { ComponentProps, forwardRef } from 'react';
 export const buttonTailwindClasses =
```
Pertempto: I think we want to remove this, now that it is no longer referenced anywhere.

Pertempto (changes requested):
Roles don't seem to be applied correctly. I create a user with Translator role, and they get stuck on the loading indicator when they try to access the translate page. I add the Admin role, and they get the same result, with no option to manage the language either.

arrocke:
@Pertempto I hadn't merged main which had the changes from the other PR which cleaned up the access control. Can you try again?

Pertempto:
If an admin changes their own access, the UI can get a little funky because it still uses the old access rights. The API protects everything though, and the frontend updates after a page refresh.

Pertempto:
Everything seems to work well!

Pertempto (approved):

PR #136, opened 2023-08-07
refactor: improve reference parsing

### LINKED ISSUE #75: verse reference format
Andrew has suggested that we also support the formats `book 1.1` and `book 1 1`. Presently, we only support the `:` separator for english, each language can have its own verse reference format.

## Proposed Format
What I'm wondering is if we can supply a standard format that handles more than one type of separator.

`<book> <chapter><separator><verse>`
* This will be reversed for RTL languages
* `book`: The localized book name. This can be shortened if the language supports it
* `chapter`: The chapter number using the number format of the language
* `separator`: Either `:`, `.`, `,`, or ` `
* `verse`: The verse number using the number format of the language

## Questions
Does this format work in all of the gateway languages?
What other separators do we need to support?
Can we get examples from all of the gateway languages?

Pertempto:
## What has changed

The reference parsing code was already implemented, just did some cleanup.

## Connected Issues

Closes #75 

## QA Steps

1. Try different reference formats in the verse selector
    - `Amos 4:4`
    - `Amos 4.4`
    - `Amos 4,4`
    - `Amos 4 4`
    - `عاموس 4 4`
## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/features/translation/verse-utils.ts
+++ b/packages/web/src/features/translation/verse-utils.ts
@@ -133,39 +133,20 @@ export function bookName(bookId: number, t: TFunction) {
 }
 
 /**
- * Convert a reference format string for a language to a reference matching regex for that language.
- * @param formatString The reference format string. See i18n.md for docs.
- * @returns A regex for matching references.
- */
-export function convertFormatToRegex(formatString: string): RegExp {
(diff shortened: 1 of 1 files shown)


arrocke (approved):
Nice work!

PR #137, opened 2023-08-08
feat: improve loading indicator

### LINKED ISSUE #103: Improve Loading Indicator
The loading indicator text isn't localized. We could replace it with some kind of spinning icon.

We should also make the verse selector persistent, and not disappear when loading.

Pertempto:
## What has changed

Replaced loading text with loading spinner in translation view.

## Connected Issues

Closes #103 

## QA Steps

1. Go to the translation page.
2. Click one of the arrows to switch verses.

The loading indicator should be shown. The verse selector should not disappear.

## Post-Deployment

- [ ] Nothing required


packages/web/src/app/Layout.tsx
```diff
@@ -27,7 +27,7 @@ export function Layout() {
     <Suspense fallback="loading">
       <div className="min-h-screen flex flex-col">
         <Header language={language} onLanguageChange={selectLanguage} />
-        <div className="flex-grow relative">
+        <div className="flex-grow relative flex flex-col">
```
arrocke, line 30: I'm not sure it's a good idea to put flex on this wrapper element. That locks every view to use column flex whether it makes sense or not. What is this needed for?

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -146,6 +143,15 @@ export default function TranslationView() {
...
       </ol>
+    );
+  }
+  return (
+    <div className="px-4 flex flex-grow flex-col gap-2">
+      <VerseSelector
+        verseId={params.verseId}
+        onVerseChange={(verseId) => navigate('../translate/' + verseId)}
+      />
+      {content}
```
arrocke: One trick I use when I a need a large conditional in the middle of a template like this is to use a function that is immediately called. This moves the conditional to the template instead of having to declare new variables and if statements outside the template. Up to you if you want to refactor this. I don't have a strong preference
```tsx
<parent>
{
  (() => {
    if (loading) {
      return <span />
    } else {
      return <div />
    }
  })()
}
</parent>
```

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -92,29 +95,23 @@ export default function TranslationView() {
...
-    );
-  }
-
-  const verse = verseQuery.data.data;
-  const referenceGlosses = referenceGlossesQuery.data.data;
-  const targetGlosses = targetGlossesQuery.data.data;
+  let content = (
+    <div className="flex-grow flex items-center justify-center">
+      <LoadingSpinner></LoadingSpinner>
+      <span className="sr-only">{t('loading')}</span>
```
arrocke: Can we move this screen reader text into the loading spinner component?

arrocke (changes requested):
Thanks for taking care of this. A few suggestions and then this should be good.

packages/web/src/app/Layout.tsx
Pertempto: So that the loading spinner could be vertically centered. The `flex-grow` on line 149 of `TranslationView.tsx` works with this `flex-col`

packages/web/src/features/translation/TranslationView.tsx
Pertempto: Done

packages/web/src/features/translation/TranslationView.tsx
Pertempto: Done

arrocke (approved):

PR #140, opened 2023-08-11
feat: update profile

### LINKED ISSUE #71: update profile
## acceptance criteria
* As a user, I should be able to update my profile from the navigation menu
* I can change my name
* If I change my email address, it doesn't get changed until I verify it
* I can change my password after confirming my current one
* I get an email when my password changed.

arrocke:
## What has changed

* Added ability to update the email, name, and password for a user
* Changing a password or email sends a notification email
* Changing an email doesn't go into affect until it is verified by clicking a link in the email

## Connected Issues

closes #71

## QA Steps

1. Login as a user
2. Change your name
3. Check that the name changed in the header
4. Change your password
5. Logout and try to log in with the new password
6. Check that an email was sent
7. Change your email address
8. Go to the verification URL in the email
9. Check the profile for the new email address
10. Check for an email sent to the old email address


packages/api/prisma/schema.prisma
```diff
@@ -77,6 +79,20 @@ model AuthKey {
   @@map("UserAuthentication")
 }
 
+/// A new email for a user awaiting verification.
+model UserEmailVerification {
```
arrocke, line 83: We can't use a key for this because we need both a token and an email address, so I just made a table for this

packages/api/shared/mailer.ts
```diff
@@ -49,23 +53,31 @@ export default {
    * @throws `MissingEmailAddressError` - If the user does not have an email address.
    */
   async sendEmail(
-    { userId, subject, text, html }: EmailOptions,
+    { subject, text, html, ...options }: EmailOptions,
```
arrocke, line 56: Because we need to send an email to the previous email address of a user, we need to be able to pass the old email here after the new one is saved on the user record. This means we are going to need to separate email verification from which email a user has. I'm going to pull that into a separate PR

packages/api/shared/redirects.ts
```diff
@@ -0,0 +1,7 @@
+const baseUrl = process.env['REDIRECT_ORIGIN'];
+
+export const redirects = {
+  invite: (token: string) => `${baseUrl}/invite?token=${token}`,
+  emailVerification: (token: string) =>
+    `${baseUrl}/verify-email?token=${token}`,
```
arrocke: Rather than pass redirect urls from the frontend, we can use an env var the frontend URL and generate them on the backend

Pertempto:
You would like this reviewed, but it won't be finished until #141 is done?

packages/api/pages/api/users/[userId].ts
```diff
@@ -3,20 +3,27 @@ import { SystemRole, UpdateUserRequestBody } from '@translation/api-types';
...
       systemRoles: z
         .array(
           z.enum(Object.values(SystemRole) as [SystemRole, ...SystemRole[]])
         )
         .optional(),
     }),
-    authorize: authorize({
-      action: 'administer',
+    authorize: authorize((req) => ({
+      action: req.body.systemRoles ? 'administer' : 'update',
```
Comment (Pertempto) [inline packages/api/pages/api/users/[userId].ts:24]: Is there ever the possibility that `req.body.systemRoles` could be the empty array? In that situation, I think your condition might not work as intended.

```suggestion
      action: req.body.systemRoles != undefined ? 'administer' : 'update',
```

packages/api/pages/api/users/[userId].ts
```diff
@@ -53,6 +61,60 @@ export default createRoute<{ userId: string }>()
...
+          },
+          data: { name },
+        });
+      }
+
+      if (password || email) {
+        const key = (await auth.getAllUserKeys(user.id)).find(
+          (key) => key.providerId === 'username'
+        );
+        if (key) {
```
Comment (Pertempto) [inline packages/api/pages/api/users/[userId].ts:77]: This makes sure that only the actual user can change their  password or email, correct?

packages/web/src/public/locales/es/users.json
```diff
@@ -4,7 +4,9 @@
   "log_in": "Acceso",
   "log_out": "Cerrar sesión",
   "password": "Contraseña",
+  "profile": "Perfil",
```
Pertempto: In the other two files, you put this on line 8.

Pertempto (changes requested):
When I tried to update a user email, I got the "Could not verify email. The verification may have expired." message in the frontend, but the email was correctly updated in the backend, and I could login with the updated email.

packages/api/pages/api/users/[userId].ts
Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:77]: If so, an explanatory comment could be helpful.

Pertempto:
Our backend now sends quite a few emails. How should we approach localizing them?

packages/api/pages/api/users/[userId].ts
Comment (arrocke) [reply packages/api/pages/api/users/[userId].ts:24]: In that scenario, it means that an admin has removed all roles from a user. Right now there is only one system role of admin, so to remove it, you have to pass an empty array.

packages/api/pages/api/users/[userId].ts
Comment (arrocke) [reply packages/api/pages/api/users/[userId].ts:77]: No, this is how we find the key that contains their password so we can update it or check that the email changed before we start the verification process

arrocke:
> When I tried to update a user email, I got the "Could not verify email. The verification may have expired." message in the frontend, but the email was correctly updated in the backend, and I could login with the updated email.

I've seen this happen when I refresh the page. It can also happen because react mounts each component twice in dev mode for some reason, so the request is fired off twice. I modified the route so that the verification route can be called multiple times. This shouldn't be a problem because it has a relatively short expiration. This should allow the page to be refreshed without a confusing error.

> Our backend now sends quite a few emails. How should we approach localizing them?

I opened an issue (#144) to address this question.

packages/api/pages/api/users/[userId].ts
Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:24]: So when an empty array is passed, we still want the action to be `administer`, correct? The logic as written would evaluate to `action: 'update'` when the `systemRoles` is empty.

Comment (Pertempto) [reply packages/api/pages/api/users/[userId].ts:24]: Ok I was mistaken. I thought the empty array evaluates to `false` in a boolean context.

Pertempto (approved):
It works well!

PR #143, opened 2023-08-13
refactor: move selected language, translate error page

### LINKED ISSUE #131: error page
We need to handle the following errors:
- not found
- no access
- other unknown errors

The error page needs to be accessible and informative

### LINKED ISSUE #135: Manage Language Should Switch Context
It always feel strange to go to the manage language page for a language, then click the translate button and it goes back to a different language.

Going to the manage language page should update the `translation-language` value in local storage, so that you are now in the subset of the site related to that language. Clicking the translate button in the header from the manage language page should now take you to the translate page for that language.

arrocke:
## What has changed

* moved the translation language selector to the translation view next to the verse selector
* add translations to the error view for not found errors and unknown error. When users access pages they don't have access to we show a not found error rather than no access
* add the header and footer to the error view
* add users to header
* redirect to translate view from the home view until we have a better home view

## Connected Issues

closes #135, closes #131 

### Selected Language
1. Log in and click translate
2. Verify that browser navigates to Gen 1:1 of Genesis
3. Change language to Spanish, verify that language changes
4. Navigate to languages list view
5. Click Translate and verify that spanish reloads, not english

### Error page
1. As an admin, enter `/languages/il` into the URL bar after the domain
2. Verify that the error page shows not found error
3. Now log out and enter `/languages/en` into the URL bar after the domain
4. Verify that the error page shows not found error
5. Now enter `/random` into the URL bar after the domain
6. Verify that the error page shows not found error

## Post-Deployment

- [x] Nothing required


packages/web/src/features/translation/router.tsx
```diff
@@ -1,15 +1,16 @@
...
-      }
-      return null;
+    path: 'translate',
+    loader() {
+      const language = localStorage.getItem('translation-language') ?? 'en';
+      return redirect(`/languages/${language}/verses/01001001`);
     },
+  },
+  {
+    path: 'languages/:language/verses/:verseId?',
```
Pertempto, line 13: Haha, you already did what I suggested [here](https://github.com/arrocke/gloss-translation/issues/135#issuecomment-1676932243).

Do we really still need `translation-language` in local storage? I know it makes it possible to implement the shortcut path `translate`, but what is the real purpose of having that path?

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -95,15 +110,27 @@ export default function TranslationView() {
...
   return (
     <div className="px-4 flex flex-grow flex-col gap-2">
-      <VerseSelector
-        verseId={params.verseId}
-        onVerseChange={(verseId) => navigate('../translate/' + verseId)}
-      />
+      <div className="flex gap-8 items-center">
+        <VerseSelector
+          verseId={verseId}
+          onVerseChange={(verseId) => navigate('../translate/' + verseId)}
```
Pertempto: This line isn't working correctly.

packages/web/src/app/Header.tsx
```diff
@@ -1,85 +1,47 @@
...
-                  >
-                    {language.name}
-                  </DropdownMenuButton>
-                ))}
-              </DropdownMenu>
-            )}
-          </>
+        <Link
+          to={'/translate'}
+          className="me-4 focus:outline-none hover:underline focus:underline"
```
Pertempto: What is the benefit of `me-4` repeatedly applied, rather than just `gap-4` on the parent? To me it seems like a lot of duplicate code.

Pertempto (changes requested):
The verse selector doesn't work at all with the router changes. Try by entering something in the input and pressing ENTER, or using the next/previous buttons.

I really like the error page and header improvements!

packages/web/src/features/translation/router.tsx
arrocke: The value for me is that by default, translate will go to english, but let's say I work on arabic 99% of the time, this will remember that for the next time.

Pertempto (approved):
Looks good!

PR #146, opened 2023-08-15
fix: move nx dependencies to dev

arrocke:
## What has changed

* moved a few dependencies to devDependencies
* prevented a few dependencies from being bundled by vercel
* fixed an import in the Header component

## QA Steps

I dunno, check that vercel completed the build

## Post-Deployment

- [x] Nothing required


packages/api/next.config.js
```diff
@@ -12,6 +12,14 @@ const nextConfig = {
...
     svgr: false,
   },
+  experimental: {
+    outputFileTracingExcludes: {
+      '*': [
+        'node_modules/@swc/core-linux-x64-gnu',
+        'node_modules/@swc/core-linux-x64-musl',
+      ],
+    },
+  },
```
arrocke, line 22: This prevents vercel from including these files in the build. These are the two biggest ones that push us over the limit. I'm not sure why they are included, they are build tools

Pertempto (approved):
It looks like the preview is working again! 👍🏽

Issue #134, opened 2023-08-04
vercel prevents import request from completing


arrocke:
**Describe the bug**

Vercel has a 10-60 second runtime limit for requests. Our import request takes a few minutes to run because there is so much data.

**To Reproduce**

1. Log in as a language or platform admin
2. Go to the manage language page
3. Import glosses
4. Observe 504 timeout error

**Expected behavior**

The import should take a while, but it should complete.

Pertempto:
tomorrow morning I might have time to run a few tests on my local machine to get an idea about what fraction of the time is spent on each task (deleting glosses, network loading, parsing, adding glosses)

arrocke:
If we were to use AWS lambda for this, here is how it could work:
1. React client makes request to `/api/languages/:code/import`
2. Server invokes the lambda function which does all of the work, creates a record in the database for the import progress with a unique token and returns 202 accepted
3. Lambda updates the import db record indicating that the import is complete
4. React client polls `/api/languages/:code/import` for the status of the job which eventually will indicate that the job is complete.

```mermaid
sequenceDiagram
    Web->>API: POST /languages/:code/import
    activate API
    API->>API: create import job in database
    API->>AWS Lambda: start job in lambda
    activate AWS Lambda
    API->>Web: Return 202
    deactivate API

    Web->>API: GET /languages/:code/import
    activate API
    API->>Web: Return 200 in-progress
    deactivate API

    AWS Lambda->>AWS Lambda: mark job as complete
    deactivate AWS Lambda
    
    Web->>API: GET /languages/:code/import
    activate API
    API->>Web: Return 200 done
    deactivate API
```

Pertempto:
It would be great if the record in the database should include some kind of progress value. Then when the frontend polls the backend, it can at least get a progress value to show to the user, so that they can see how much of the import is complete.

Why do we need the `/languages/:code/import/complete` API route? Why can't the lambda just update the database record to indicate that the import is complete?

Pertempto:
In my rough tests, the various parts of import take up this much of the total import time:
| Task | Fraction |
|-|-|
| Delete old glosses | ~30% |
| Fetch data | ~28%   |
| Parse data |  ~2%  |
| Insert new glosses |  ~39%  |
| Other processing  | ~0.4%   |


[...]

PR #148, opened 2023-08-18
refactor: extract db to separate package

arrocke:
## What has changed

* New `@translation/db` package that exports prisma types and a prisma client
* I changed the location of the client to use the default which generates the client as a node module (`@prisma/client`). The other PR/branch generates the client in the package, which would allow us to support multiple prisma clients, but that is blocked by https://github.com/prisma/prisma/issues/19784

## Post-Deployment

- [x] Nothing required


packages/db/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from '@prisma/client';
```
arrocke, line 1: For now, this package just exports the prisma client. Doing it this way will allow us to eventually bundle the client in this package so we could support multiple prisma schemas.

packages/api/shared/db.ts
```diff
@@ -1,8 +1,8 @@
-import { PrismaClient, Prisma } from '../prisma/client';
+import { PrismaClient, Prisma } from '@translation/db';
 import { randomBytes } from 'crypto';
 
 export { Prisma };
-export * as PrismaTypes from '../prisma/client';
+export * as PrismaTypes from '@translation/db';
 
 export const client = new PrismaClient();
```
arrocke, line 7: Now when we need to connect to the database, we import from `@translation/db` rather than `@prisma/client`. Same thing for prisma types.

docs/db.md
```diff
@@ -2,28 +2,28 @@
...
+[Database schema](./../packages/db/prisma/schema.prisma)
 
 ## Setup
 
 1. Install postgresql version 14.7
 1. Create database for the project
-1. Set the `DATABASE_URL` env var in `packages/api/.env.local` to the postgres connection string. You may need to include a username and password in the connection string.
-1. Run `nx prisma api migrate reset` to scaffold the database schema.
-1. Run `nx seed api` to insert static verse data into the database.
+1. Set the `DATABASE_URL` env var in `packages/api/.env.local` and `packages/db/.env.local` to the postgres connection string. You may need to include a username and password in the connection string.
```
Pertempto, line 11: Do we still need it in `packages/api/.env.local`?

.eslintignore
```diff
@@ -1,2 +1,2 @@
 node_modules
-packages/api/prisma/client
\ No newline at end of file
+packages/db/src/client
```
Pertempto: This directory doesn't exist.

Pertempto (changes requested):
Hmmm, I'm not sure what's going on

1. `npm i` runs fine.
1. I ran `nx serve api` and `nx serve web`
1. I tried to access `http://localhost:4200/languages/en/verses/01001001`, but all the API requests hang indefinitely. 

Did I miss a setup step?

docs/db.md
arrocke: Yep, the one in the db package doesn't get loaded when the api project is served

arrocke:
Did you see any error messages in the api logs?

Pertempto:
> Did you see any error messages in the api logs?

I manually went to the login page, waited a while for everything to load after I submitted username and password, and it finally loaded.

It did get this error in the API logs:
```
Invalid `prisma.authSession.delete()` invocation:
An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
    at Zr.handleRequestError (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\@prisma\client\runtime\library.js:171:6414)
    at Zr.handleAndLogRequestError (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\@prisma\client\runtime\library.js:171:5948)
    at Zr.request (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\@prisma\client\runtime\library.js:171:5786)
    at async t._request (C:\Users\addis\SoftwareProjects\gloss-translation\gloss-translation\node_modules\@prisma\client\runtime\library.js:174:10455)
    at async Object.deleteSession (file:///C:/Users/addis/SoftwareProjects/gloss-translation/gloss-translation/node_modules/@lucia-auth/adapter-prisma/index.js:100:13) {
  digest: undefined
}
```

Pertempto:
Otherwise, everything seems to load. It almost seems like it takes longer to load glosses. It is taking 3+ seconds to load glosses for each verse

image


Hopefully this doesn't occur on our production server, or the UX quality will really drop.

Pertempto (approved):

Pertempto:
I think we should go ahead and merge this, so I can keep going on #134.

If slow load times is a problem on production, we can create a separate issue for it.

PR #151, opened 2023-08-23
refactor: use db package from lambda function

Pertempto:
## What has changed

Successfully hooked up our `lambda-functions` package with our `db` package.

## Connected Issues

Related to #134 
## QA Steps

1. Follow the setup steps in the `Lambda Functions` section of `docs/contributing.md`
2. Access `localhost:3000/hello`
    - If you have an `ar` language in your database, you should get the name of that language.
    - Otherwise, you should get a message "Not found"

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -27,7 +27,7 @@ These are the major technologies we are using. Familiarity with most of these wi
 
 1. [Install](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-sam-cli.html#install-sam-cli-instructions) the SAM cli.
 1. [Install](https://www.docker.com) docker.
-1. Copy `env-example.json` to `env.json` to the lambda-functions project and update with your local env vars. You can use `host.docker.internal` in situtations where the host IP address is needed. This allows the docker container to connect to services on your host machine.
+1. Copy `env-example.json` to `env.json` to the lambda-functions project and update with your local env vars. You can use `host.docker.internal` in situations where the host IP address (AKA `localhost`) is needed. This allows the docker container to connect to services on your host machine.
 1. Run `nx serve lambda-functions` in a terminal to make the lambda functions available on port 3000.
    - **Note** - you have to restart this every time you make code changes. Hot reload isn't trivial when running local lambda functions - see [here](https://github.com/aws/aws-sam-cli/issues/901) and [here](https://github.com/aws/aws-sam-cli/issues/921).
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -27,7 +27,8 @@ These are the major technologies we are using. Familiarity with most of these wi
 
 1. [Install](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-sam-cli.html#install-sam-cli-instructions) the SAM cli.
 1. [Install](https://www.docker.com) docker.
-1. Run `nx serve-functions lambda-functions` in a terminal to make the lambda functions available on port 3000.
+1. Copy `env-example.json` to `env.json` to the lambda-functions project and update with your local env vars. You can use `host.docker.internal` in situtations where the host IP address is needed. This allows the docker container to connect to services on your host machine.
+1. Run `nx serve lambda-functions` in a terminal to make the lambda functions available on port 3000.
    - **Note** - you have to restart this every time you make code changes. Hot reload isn't trivial when running local lambda functions - see [here](https://github.com/aws/aws-sam-cli/issues/901) and [here](https://github.com/aws/aws-sam-cli/issues/921).
 
 ## Development Environment
--- a/packages/lambda-functions/env-example.json
+++ b/packages/lambda-functions/env-example.json
@@ -1,0 +1,5 @@
+{
+  "HelloWorldFunction": {
+    "DATABASE_URL": ""
+  }
+}
--- a/packages/lambda-functions/functions/hello-world/app.ts
+++ b/packages/lambda-functions/functions/hello-world/app.ts
@@ -1,22 +1,117 @@
+import { PrismaClient } from '@translation/db';
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
+import { bookKeys } from '../../../../data/book-keys';
 
-/**
- *
- * Event doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
- * @param {Object} event - API Gateway Lambda Proxy Input Format
- *
- * Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
- * @returns {Object} object - API Gateway Lambda Proxy Output Format
- *
- */
+const client = new PrismaClient();
+
+const importServer = 'https://hebrewgreekbible.online';
 
 export const lambdaHandler = async (
   event: APIGatewayProxyEvent
 ): Promise<APIGatewayProxyResult> => {
+  console.log(event.queryStringParameters);
   return {
     statusCode: 200,
     body: JSON.stringify({
       message: 'Wow it actually works!',
     }),
   };
+
+  const languageCode = 'ar'; // TODO: read language code from request
+  const importLanguage = 'Arabic'; // TODO: read import language from request
+
+  const language = await client.language.findUnique({
+    where: {
+      code: languageCode,
+    },
+  });
+
+  // Delete all the glosses for the language.
+  await client.gloss.deleteMany({
+    where: {
+      languageId: language.id,
+    },
+  });
+
+  const glossData: { wordId; gloss }[] = [];
+  for (const key of bookKeys) {
+    const bookId = bookKeys.indexOf(key) + 1;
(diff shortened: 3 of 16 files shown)


Pertempto:
I'm having issues getting the lambda function to build and import the `@translation/db` package. I'll keep trying.

Pertempto:
@arrocke Should we have two API routes for this? A POST route to trigger the job, a GET route to poll the progress.

arrocke:
> Should we have two API routes for this? A POST route to trigger the job, a GET route to poll the progress.

Yeah, that's what I was thinking in #134

Pertempto:
> Yeah, that's what I was thinking in #134

Ah I see it in the chart now 👍🏽

Pertempto:
@arrocke From what I can tell so far, nx and lambda do not mix well at all. This might be past my skill level.

Pertempto:
I can't figure out the correct way to get lambda to recognize our nx packages and import them. I'm not finding any great examples/docs online either.

packages/lambda-functions/functions/hello-world/app.ts
```diff
@@ -1,22 +1,117 @@
+import { PrismaClient } from '../../../../packages/db/src/index';
```
Pertempto: Lambda wasn't recognizing the `@translation/db` alias from `tsconfig.base.json`, so I used this instead.

But now I get this error when I try to run the lambda function: `ENOENT: no such file or directory, open '/var/task/schema.prisma'`

I think this is what happens when it tries to access the prisma schema from inside the docker container.

packages/lambda-functions/project.json
```diff
@@ -4,17 +4,43 @@
...
+            "glob": "schema.prisma",
+            "output": "/"
+          },
+          {
+            "input": "node_modules/.prisma/client",
+            "glob": "libquery_engine-rhel-openssl-*.so.node",
+            "output": "/"
+          }
+        ]
       }
```
arrocke, line 29: I'm using webpack to build the lambda. This gets us two important benefits:
* This target is dependent on the `db` package, so nx will build that first.
* We can include the prisma schema and the engine shared library as assets and webpack will include them in the build.

packages/lambda-functions/project.json
```diff
@@ -4,17 +4,43 @@
...
     },
-    "deploy-functions": {
-      "command": "sam.cmd build && sam.cmd deploy --guided",
+    "serve": {
+      "command": "sam local start-api --env-vars env.json",
       "options": {
         "cwd": "packages/lambda-functions"
-      }
+      },
+      "dependsOn": ["build"]
```
arrocke, line 36: Now before we can serve this locally, we can tell nx to build it first

packages/lambda-functions/template.yaml
```diff
@@ -12,27 +12,20 @@ Resources:
   HelloWorldFunction:
     Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
     Properties:
-      CodeUri: functions/hello-world
-      Handler: app.lambdaHandler
+      CodeUri: ../../dist/packages/lambda-functions
```
arrocke, line 15: We can serve and deploy this function from the dist folder now that it has been built by webpack. The contents of this folder are what will be deployed to lambda

packages/lambda-functions/template.yaml
```diff
@@ -12,27 +12,20 @@ Resources:
...
+      Handler: main.lambdaHandler
       Runtime: nodejs18.x
       Architectures:
         - x86_64
+      Environment:
+        Variables:
+          DATABASE_URL:
```
arrocke, line 22: We also need to define an environment variable for the database url

packages/lambda-functions/env-example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "HelloWorldFunction": {
+    "DATABASE_URL": ""
+  }
+}
```
arrocke, line 5: If you copy this file to `env.json` and set your connection string using `host.docker.internal` instead of localhost, this file will be read by SAM to set the environment of the lambda in the docker container

arrocke (commented):
Hey, I got it to work! Can you pull this down and confirm it works on your machine? I'll clean the rest of this up if that works as expected

Pertempto:
> Hey, I got it to work! Can you pull this down and confirm it works on your machine? I'll clean the rest of this up if that works as expected

Awesome! I'm busy this evening, but I'll test in the morning.

docs/contributing.md
```diff
@@ -27,7 +27,8 @@ These are the major technologies we are using. Familiarity with most of these wi
 
 1. [Install](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-sam-cli.html#install-sam-cli-instructions) the SAM cli.
 1. [Install](https://www.docker.com) docker.
-1. Run `nx serve-functions lambda-functions` in a terminal to make the lambda functions available on port 3000.
+1. Copy `env-example.json` to `env.json` to the lambda-functions project and update with your local env vars. You can use `host.docker.internal` in situtations where the host IP address is needed. This allows the docker container to connect to services on your host machine.
```
Pertempto:
```suggestion
1. Copy `env-example.json` to `env.json` to the lambda-functions project and update with your local env vars. You can use `host.docker.internal` in situations where the host IP address (AKA `localhost`) is needed. This allows the docker container to connect to services on your host machine.
```

Pertempto:
Do you get this during the build section of  `nx serve lambda-functions`?

image

Pertempto:
I created my `env.json` file in `lambda-functions` and re-ran `npm i`, but I still get errors in `nx serve lambda-functions`

It appears the schema file isn't copied over correctly.
image

Pertempto:
For some reason I still have `db-client` and `db` directories. I should only have `db`, right?

packages/lambda-functions/template.yaml
Pertempto: Do I need to set a value here in testing? Or does that get set by `env.json`

packages/lambda-functions/project.json
```diff
@@ -4,17 +4,43 @@
...
+        "target": "node",
+        "externalDependencies": "none",
+        "outputPath": "dist/packages/lambda-functions",
+        "main": "packages/lambda-functions/functions/hello-world/app.ts",
+        "tsConfig": "packages/lambda-functions/tsconfig.json",
+        "assets": [
+          {
+            "input": "packages/db/src/",
+            "glob": "schema.prisma",
+            "output": "/"
```
Pertempto, line 21: I tried updating this to `/var/task/` to see if that would resolve the error I was seeing, but when I check the docker files there is only `app.js` in `/var/task`

arrocke: `/var/task` is a directory within the docker container where the source code from outside the container is mounted. That's why you aren't seeing anything different when you do this. If you have a /var/task directory on your computer, that's where this went, but the docker container can't see that.

arrocke:
> Do you get this during the build section of nx serve lambda-functions?

Yep, I'm seeing that. I'm not sure what we can do about it if the prisma client package doesn't have source maps.

> I created my env.json file in lambda-functions and re-ran npm i, but I still get errors in nx serve lambda-functions

In the error log it says that it is mounting a directory in `.aws-sam`. That's going to be a problem because that was the directory that sam builds to when it was using esbuild. Now that we are using webpack, it needs to mount `dist/packages/lambda-functions`. So I would confirm that there are the correct files in that directory, and delete the `.aws-sam` folder to see if that resets everything.

> For some reason I still have db-client and db directories. I should only have db, right?

You can delete the db-client directories. When git changes branches, it doesn't delete directories for some reason.

Pertempto:
Deleting `.aws-sam` did the trick. It works! :tada:

Thanks a ton for figuring this out!

packages/lambda-functions/template.yaml
arrocke: This is just so that sam knows what fields to read from env.json

Pertempto:
@arrocke Do you have any clean up to finish on this branch? In order to keep PR sizes down, I think we should go ahead and merge this, and then I'll keep building the import logic on top.

packages/lambda-functions/functions/hello-world/app.ts
```diff
@@ -1,22 +1,120 @@
...
- *
- * Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
- * @returns {Object} object - API Gateway Lambda Proxy Output Format
- *
- */
+const client = new PrismaClient();
+
+const importServer = 'https://hebrewgreekbible.online';
 
 export const lambdaHandler = async (
```
Pertempto, line 9: How will we lock this down and make sure that not anyone can trigger an import process?

We can probably setup the AWS stuff so that only our API has permission to access the lambda functions?

Pertempto:
@arrocke Please take a look and merge if this looks good for now. I cleaned up the API routes/types a bit to prepare for the actual logic.

packages/lambda-functions/functions/hello-world/app.ts
arrocke: Yeah, I think that's the easiest approach

arrocke (approved):
I'll merge what we have and then keep working in a separate branch

PR #191, opened 2023-09-06
feat: Persist Translation Verse ID

### LINKED ISSUE #167: save reference when returning to translation view
**Describe the bug**
The reference is forgotten when switching tasks. The language of translation is remembered but the page goes back to Genesis 1:1.

**To Reproduce**

1. Navigate to Gen 1:2 or any reference other than Gen 1:1
2. Click on 'Languages'
3. Click on 'Translate'
4. Goes back to Gen 1:1

**Expected behavior**
I would expect it to return to the last reference that was being worked on or viewed.

**Screenshots**
Not necessary.

- Device: Desktop
- OS: Windows 11
- Browser: Edge
- Version: ???

**Additional context**

Pertempto:
## What has changed

Just like we persist the translation language when you leave the translation page, now we also persist the verse ID.

## Connected Issues

#167 

## QA Steps

1. Navigate to Gen 1:2 or any reference other than Gen 1:1
1. Click on 'Languages'
1. Click on 'Translate'
    - It should go back to the verse you were previously on

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslationView.tsx
+++ b/packages/web/src/features/translation/TranslationView.tsx
@@ -12,16 +12,24 @@ import DropdownMenu, {
   DropdownMenuLink,
 } from '../../shared/components/actions/DropdownMenu';
 
+export const translationLanguageKey = 'translation-language';
+export const translationVerseIdKey = 'translation-verse-id';
+
 export default function TranslationView() {
-  const { verseId, language } = useParams() as {
-    verseId: string;
+  const { language, verseId } = useParams() as {
     language: string;
+    verseId: string;
   };
-  const navigate = useNavigate();
 
   useEffect(() => {
-    localStorage.setItem('translation-language', language);
+    localStorage.setItem(translationLanguageKey, language);
   }, [language]);
+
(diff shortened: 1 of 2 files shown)


packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -12,17 +12,25 @@ import DropdownMenu, {
   DropdownMenuLink,
 } from '../../shared/components/actions/DropdownMenu';
 
+export const translationLanguageKey = 'translation-language';
+export const translationVerseIdKey = 'translation-verse-id';
+
 export default function TranslationView() {
-  const { verseId, language } = useParams() as {
-    verseId: string;
+  const { language, verseId } = useParams() as {
```
Pertempto, line 19: I re-ordered these just to match the order they are specified in the paths. That way we can have consistency in all the code I touched: the language is read, and then the verse ID.

Pertempto:
This was a simple improvement! 😄

arrocke (approved):
Nice work! This looks good

PR #193, opened 2023-09-06
refactor: up and down arrow directions for verse selector

### LINKED ISSUE #181: Arrow direction to change verses
To avoid confusion, it would be better to have the arrows go up and down rather than left right. So that no matter which direction the language is written, it'll be clear which arrow goes to the next verse.

arrocke:
## What has changed

* Switched to up and down arrows for the verse selector buttons. That makes it clearer which way the verse changes regardless of language orientation

## Connected Issues

closes #181 

## QA Steps

1. Go to the translation view
2. Use the arrow buttons
3. Confirm verse changes as you'd expect

## Post-Deployment

- [x] Nothing required


Pertempto (approved):
It works well and makes sense on both LTR and RTL layouts.

PR #194, opened 2023-09-06
refactor: make chapter and verse optional when parsing reference

### LINKED ISSUE #185: Navigating to a reference with chapter but no verse number
When I type in Psalm 140 without a verse number into the reference navigator it simply doesn't do anything. It would be nice if it could just simply go to the first verse of that psalm automatically.

arrocke:
## What has changed

verse and chapter numbers are now optional when you type in a reference. It will fall back to chapter 1 and verse 1 if not provided.

## Connected Issues
closes #185 

## QA Steps

1. In the translation view, try typing Psalm 100 and Obadiah.
2. In both cases it should fill in the chapter and verse number that are missing

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/verse-utils.ts
+++ b/packages/web/src/features/translation/verse-utils.ts
@@ -139,7 +139,7 @@ export function bookName(bookId: number, t: TFunction) {
  * @returns The verse ID if it can be determined, otherwise `null`.
  */
 export function parseReference(reference: string, t: TFunction): string | null {
-  const referenceRegex = /^(.+)\s+(\d+)([:.,]|\s+)(\d+)$/;
+  const referenceRegex = /^(.+?)(:?\s+(\d+)(?:([:.,]|\s+)(\d+))?)?$/;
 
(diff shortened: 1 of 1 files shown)


Pertempto (changes requested):
I found a way to trick it! 😁 

Try "obd 2:34". On production this goes to `31001021`

Here it goes to `31001NaN`.

Also one small note: when we changed "Psalms" to "Psalm" for the verse references (which I think was a good idea), now people can't enter "Psalms" to jump to "Psalm 1:1". I'd think "Psalms" is the name of the book, even if you use "Psalm" to refer to a particular psalm in the psalms.

This is by no means a big deal, it might just cause very rare occasions of confusion with the users. Maybe we could make our book matching logic a little fuzzier, so that it still matches with an extra letter or two at the end of a book name?

arrocke:
Lol, my regex had a typo. Should be fixed now

Pertempto (approved):

PR #195, opened 2023-09-06
refactor: use sbl fonts for greek and hebrew

### LINKED ISSUE #187: Different Hebrew Font
The Hebrew for on the site needs to change to something other than Times New Roman. Times New Roman doesn't display _furtive patachs_ correctly. I think we had already talked about only using the SBL font.

arrocke:
## What has changed

Added SBL fonts for hebrew and greek using tailwind utilities `font-hebrew` and `font-greek`

## Connected Issues

closes #187 

## QA Steps

Check out a verse in the NT and OT to compare the new fonts to what is in production

## Post-Deployment

- [x] Nothing required


Pertempto (approved):
Wow this looks so much better! Even an untrained eye like mine can tell how much nicer these fonts are. 😆

PR #196, opened 2023-09-07
feat: add lemma resource data structure

### LINKED ISSUE #162: lexicon data structure
* We need a new table that contains lexicon data for each lemma
* Primary key will combine the lemma id and the lexicon code (`bdb`, `lsj`, `strongs`)
* Table needs a string field with the markdown entry.

Pertempto:
## What has changed

Added `LemmaResource`

## Connected Issues

Closes #162 

## QA Steps

- Run `npm i`
    - There should be no errors
- Run `nx prisma db migrate dev`
    - There should be no errors

## Post-Deployment

- [x] Nothing required


packages/db/src/schema.prisma
```diff
@@ -265,3 +267,24 @@ model Gloss {
 
   @@id([wordId, languageId])
 }
+
+/// The entry in a lexicon associated with a particular lemma.
+/// TODO: determine if we need a better name
+model LemmaLexiconEntry {
```
Pertempto: I wasn't sure what to name it.

Pertempto:
As stated before, this is my first time updating the schema. I'm sure I made some mistakes.

packages/db/src/schema.prisma
arrocke: I can think of two alternatives, but I don't have a strong preference
1. LexiconEntry - Lexicon already communicates that we are talking about dictionary like entries.
2. LemmaResource - Technically, Strongs isn't a lexicon, but a concordance. If we call these resources for lemmas, then a LemmaResource would have a resourceCode instead. Also, a Lemma would have resources rather than lexiconEntries

arrocke (approved):

Pertempto:
I like `LemmaResource` I'll update this.

Pertempto:
@arrocke Just to be extra sure, can you double check that my migration and schema updates work correctly before merging into the main branch?

arrocke:
Yep, everything came through just fine!

PR #197, opened 2023-09-08
refactor: prefetch the next 3 verses while translating

### LINKED ISSUE #186: Lag when navigating to new verse
When going to a new verse there's a significant lag (The lag isn't very significant when going back to a verse that has already loaded). It doesn't really make sense for loading one verse to take that long. Sheldon's site is super snappy in comparison, and I think that would be a nice standard to shoot for.  
Chrome browser.

arrocke:
## What has changed

When the translation view loads for each verse, it primes the cache with the next three verses if they haven't been loaded.

## Connected Issues

closes #186 

## QA Steps

1. Start translating a verse
2. Click the next verse button
3. Confirm that there is no network latency when loading the next verse

## Post-Deployment

- [x] Nothing required


Pertempto (approved):
This seems to work well. Rapidly clicking the next button is obviously going to cause the buffer of 3 to be not enough, but that shouldn't really be a normal use case.

PR #198, opened 2023-09-09
feat: basic navigation shortcuts

### LINKED ISSUE #188: Shortcut keys for navigation
BART has a good system of shortcut keys I'd like to keep. People are already used to them, so if we're going to get people to switch over, they'll need to have the shortcuts they're used to.

Pertempto:
## What has changed

Added 6 basic navigation shortcuts

## Connected Issues

Related to #188 

## QA Steps

Test the following shortcuts:

ctrl up/down keys - go to prev/next verse respectively
ctrl home - go to first word of the current verse
ctrl end - go to last word of the current verse
ctrl shift home - go to first verse of book
ctrl shift end - go to last verse of book

## Post-Deployment

- [ ] Nothing required


packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -112,6 +116,59 @@ export default function TranslationView() {
...
+          } else {
+            console.log('FIRST WORD!');
+          }
+          break;
+        case 'End':
+          if (event.shiftKey) {
+            console.log('LAST VERSE!');
+          } else {
+            console.log('LAST WORD!');
+          }
```
Pertempto: I've got previous/next verse functionality working, I still need to implement the other 4 shortcuts. The first/last verse of book functionality should be easy, I just need to write the helper function that calculated the verse ID.

@arrocke Any hints for how to implement the first/last word shortcuts? How would I manually move focus to the correct `TranslateWord` components?

Pertempto: This morning I learned how to use `forwardRef`! :smile:

Pertempto:
I noticed that the translation view doesn't automatically focus the first word after the verse loads.

Where would be the correct place to place `firstWord.current?.focus();` to implement that functionality?

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -146,6 +152,66 @@ export default function TranslationView() {
 
   const userCan = useAccessControl();
 
+  const firstWord = useRef<HTMLInputElement>(null);
+  const lastWord = useRef<HTMLInputElement>(null);
+  const handleKeyPress = useCallback(
+    (event: {
```
arrocke, line 158: this can be typed as `KeyboardEvent`

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -15,91 +15,106 @@ export interface TranslateWordProps {
...
-  status,
-  gloss,
-  referenceGloss,
-  previousGlosses,
-  onGlossChange,
-}: TranslateWordProps) {
-  const { t } = useTranslation();
-  const [text, setText] = useState(gloss ?? '');
-  const width = useTextWidth(text);
+const TranslateWord = forwardRef<HTMLInputElement, TranslateWordProps>(
```
arrocke: This component represents a group of controls, rather than a single input, so I think we should have its ref be our own interface. We can call it `TranslateWordRef` and have it just provide a focus function. We don't really want to expose the actual input element because that's an individual element of this component. This is where the hook `useImperativeHandle` lets us expose a custom object as the ref.

That's contrasted with the autocomplete input where the input element better represents the component and all of the other stuff augments it to make the input more powerful.

arrocke (changes requested):
Control and arrow keys conflict with OS shortcuts on mac. I don't want to block this PR with that since we don't have a solution, so I'll mention it in the issue

packages/web/src/features/translation/TranslationView.tsx
Pertempto: That causes errors, because I use this callback from two different places, which use two different types called `KeyboardEvent`. The one place calls it with the builtin `KeyboardEvent`, the other place calls it with the react type `KeyboardEvent<HTMLElement>`. This type I supplied here contains just the properties that I'm interested in, which the two types share.

The react `KeyboardEvent` seems to conflict with the builtin [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)

packages/web/src/features/translation/TranslateWord.tsx
Pertempto: Improved

packages/web/src/features/translation/TranslationView.tsx
arrocke: Ah, that makes more sense

arrocke (approved):
Just realized that I forgot to respond to your comment about focusing the first word in the verse. I think we can leave that out for now. I'm still thinking about how we want that to work. It might be a problem if the initial focus element is hijacked, so we might need to only focus in certain situations.

Pertempto:
> Just realized that I forgot to respond to your comment about focusing the first word in the verse. I think we can leave that out for now. I'm still thinking about how we want that to work. It might be a problem if the initial focus element is hijacked, so we might need to only focus in certain situations.

I'm confused. I successfully implemented the first/last word shortcuts. Do you want me to take them out now?

arrocke:
I was referring to this comment: https://github.com/arrocke/gloss-translation/pull/198#issuecomment-1713478369

PR #205, opened 2023-09-18
refactor: improve i18n organization

### LINKED ISSUE #142: localization strategy
After using react-i18next a bit, I think we can improve a few things.

* Create a `common` namespace for keys that could be used on any view. For example: `Submit`, `Next`, `Back`
* Create new namespaces for keys that are limited to a subdomain of the app. For example: the users namespace could have the keys for `Email` and `Password`, but perhaps the the key for `Name` is in the common namespace since that can be used by languages.
* Require components to use the full namespace when specifying translation keys. That makes it easier to add a namespace later since no keys are assuming a default namespace.
* Sort the translation keys in alphabetical order

Pertempto:
## What has changed

Separated out our translation terms into separate files, depending on the UI context in which they are used.

Updated all the usages of `t()` and `useTranslate()` that didn't include the namespace name.

## Connected Issues

Closes #142 

## QA Steps

- Go to the different pages of the UI
    - Try in all three languages, look for un-translated terms.
        - It can be tricky to spot un-translated terms if you are looking in English, but opening the dev tools console will show any translate term miss errors

## Post-Deployment

- [ ] Nothing required


Pertempto:
@arrocke there is still quite a bit of translation term cleanup/normalization we could do, but I tried to avoid going down that rabbit trail in this issue. This PR is big enough, we can work on that in a future issue/PR.

arrocke (approved):
This is great! Thanks for taking care of this

PR #206, opened 2023-09-19
feat: language font selector

### LINKED ISSUE #168: Gloss font
When I work in Urdu as a gloss language the text is difficult to read.

I would like to see the ability to display the gloss text in another font such as Scheherazade that has been tweaked for Urdu or better yet, an Urdu Nastaliq font such as Mehr.

In the Languages tab for each language any admin would have the ability to upload a font and choose to use that font for displaying the language.

**Additional context**
none.

Pertempto:
## What has changed

- Added gloss font selector to the manage language page.
- Use gloss font on translate page.
- By default, new languages use "Noto Serif" as their gloss font

## Connected Issues

Resolves #168

## QA Steps

Be sure to set `NX_GOOGLE_FONT_API_KEY` in `.env.local` to your google font API key.

### Default Gloss Font
The inputs on the translate page should by default be using "Noto Serif"

### New Language
1. Create a new language
2. Go to the manage language page for that language.
The selected gloss font should be "Noto Serif"

### Change Gloss Font
1. Go to the manage language page for a language
2. Select a font from the gloss font selector
3. Press the update button
4. Reload the page
    - The selected font setting should persist
5. Go to the translate page for that language
    - The gloss inputs should use the gloss font

## Post-Deployment

- [x] Setup the `NX_GOOGLE_FONT_API_KEY` environment variable for the web build.


packages/web/src/shared/fontClient.ts
```diff
@@ -0,0 +1,33 @@
+const FONT_API =
+  'https://www.googleapis.com/webfonts/v1/webfonts?key=' +
+  process.env.GOOGLE_FONT_API_KEY;
+
+class FontClient {
+  async getFonts() {
+    console.log(JSON.stringify(process.env));
+    console.log(process.env.API_URL);
+    console.log(process.env.GOOGLE_FONT_API_KEY);
```
Pertempto: @arrocke, for some reason `process.env.GOOGLE_FONT_API_KEY` is always `undefined`. I defined it in `packages/web/.env.local`, but it doesn't seem to register. Is there anything special I need to do to get it to work?

packages/web/src/features/languages/ManageLanguageView.tsx
```diff
@@ -108,6 +110,8 @@ export default function ManageLanguageView() {
 
   const { data: language } = useLicenseQuery(params.code);
   const { data: members } = useLicenseMembersQuery(params.code);
```
Pertempto: @arrocke What is the `useLicense` all about? It's confusing me. Is `useLoaderData` an appropriate way to load my list of font options?

packages/web/src/shared/fontClient.ts
arrocke: You might need to restart your web server for nx to pick up the new vars

packages/web/src/features/languages/ManageLanguageView.tsx
arrocke: Lol. That function should be named `useLanguageQuery`. At my job, we have licenses as our top level org unit rather than languages. Probably not the last time I'll interchange them.

`useLoaderData` is the pattern for react router for loading data before the navigation completes. That should work great in this case. The new view won't render until all of the data has been fetched

packages/web/src/shared/fontClient.ts
Pertempto: It's really strange. I've tried that several times and it did not fix it. I also tried directly adding it to `.env` and restarting but that didn't work either.

Pertempto: I'll try adding an `NX_` prefix: https://stackoverflow.com/a/71063731

Pertempto: Ok, that worked. Not sure how `API_URL` works?

Pertempto:
@arrocke 
How should I filter down the list of fonts? Only sans fonts? Only Noto fonts?

Loading 1567 fonts for previews in the font picker isn't feasible

packages/web/src/shared/fontClient.ts
arrocke: API_URL works because of [this file](https://github.com/arrocke/gloss-translation/blob/main/packages/web/webpack.config.js) which configures webpack to replace references to API_URL with that string. Let's stick with the NX_ prefix so that we don't have modify our tooling for every env var

arrocke:
I think we should limit to Noto fonts for now. Most of the other fonts are going to be just noise for users to sort through.

A couple ideas for rendering the list. In all cases, we can load the full font after selection to show a preview to help the user confirm what they are expecting.
* Render all font options in the list with the UI font.
* Use the menu font url from the Google Font API which includes just the characters needed to render the language name.
* Precompute SVGs of each font name rendered with it's font. This is what figma does.

Pertempto:
If I use the `text` filter to the css API, the resulting woff2 file is even smaller than the ttf file pointed to by the `menu` URL

For example: https://fonts.googleapis.com/css2?family=Noto%20Nastaliq%20Urdu&text=Noto%20Nastaliq%20Urdu

Pertempto:
@arrocke I've got the basic preview UI working for the selector. Try it out!

The main problem is that the preview isn't very useful for fonts that aren't aimed at latin characters. Google fonts uses a different preview text for those fonts

image

I'm not sure how we handle that situation. We could have some preview text beside the selector that changes based on the `subsets` values returned by the API. We would need to create a mapping from subsets to useful preview texts.


I still need to do the backend work to actually persist the selected gloss font.

Pertempto:
Should I update the documentation to mention developers setting up their own Google font API key for testing?

packages/api-types/src/index.ts
```diff
@@ -82,6 +82,7 @@ export interface UpdateUserRequestBody {
 export interface Language {
   code: string;
   name: string;
+  glossFont: string;
```
arrocke: Nitpicky, but I'd prefer to use `font` instead of `glossFont`. We'll want to use this font for the reference translations as well, so making this more general is applicable

packages/db/src/schema.prisma
```diff
@@ -190,6 +190,8 @@ model Language {
   code              String               @unique
   /// The localized name of the language.
   name              String
+  /// The English name of the gloss font selected for this language.
+  glossFont         String               @default("Noto Serif")
```
arrocke: Same naming thing for the database

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -79,16 +88,27 @@ export default function TranslationView() {
...
+  useEffect(() => {
+    if (selectedLanguage) {
+      document.head.insertAdjacentHTML(
+        'beforeend',
+        `<link rel=stylesheet href="${fontClient.getCssUrl(
+          selectedLanguage.glossFont
+        )}">`
+      );
+    }
+  }, [selectedLanguage]);
```
arrocke: This works really well to load the list of fonts. Can we turn this into a reusable hook that could also check if that font URL has been loaded already?

packages/web/src/features/languages/ManageLanguageView.tsx
```diff
@@ -154,6 +172,26 @@ export default function ManageLanguageView() {
...
+          <div className="mb-2">
+            <FormLabel htmlFor="glossFont">
+              {t('languages:gloss_font').toUpperCase()}
+            </FormLabel>
+            <SelectInput
+              id="glossFont"
+              name="glossFont"
+              className="w-full h-fit min-h-[40px]"
+              required
+              value={previewFont}
```
arrocke, line 181: The initial value isn't being set to what was saved, so it is defaulting to the first font in the list.

arrocke (changes requested):
This works remarkably well! A few things to clean up.

packages/web/src/features/languages/ManageLanguageView.tsx
Pertempto: See `const [previewFont, setPreviewFont] = useState(language.data.font);` on line 136.

It's working fine for me.

packages/web/src/features/translation/TranslationView.tsx
Pertempto: Take a look at the new hook I made: `useLoadFonts`

packages/web/src/shared/hooks/useLoadFonts.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect } from 'react';
+import fontClient from '../fontClient';
+
+/**
+ * Ensure that each of the given fonts has been added to the page.
+ */
+export const useLoadFonts = (fonts: string[], preview = false) => {
```
arrocke: Another nitpick for naming. Let's call this `useFontLoader` so that we don't have two verbs in a row.

arrocke (approved):
The only other thing that needs addressed is that we need to stack fonts. Here is from Googles [docs](https://fonts.google.com/noto/use#how-are-noto-fonts-organized) about Noto

> If you have text in multiple languages that use different scripts, either the app or the person who typesets the text needs to switch between the different fonts accordingly. For example, [Noto Sans Armenian](https://fonts.google.com/noto/specimen/Noto+Sans+Armenian) only has Armenian-script letters. But Armenian text also contains digits, punctuation, it may also contain terms written in Latin or Cyrillic scripts, so you also need the main Noto Sans fonts.

So I propose that whenever we need to load and use a specific sans or serif noto font, we also need to load and use the corresponding `Noto Sans` or `Noto Serif` font as a second font.

Pertempto:
> So I propose that whenever we need to load and use a specific sans or serif noto font, we also need to load and use the corresponding Noto Sans or Noto Serif font as a second font.

You want this implemented in this PR? How would I detect that? Just use the name of the font and check if it contains "sans" or "serif"?

Pertempto:
@arrocke I attempted some basic logic for loading a secondary font. Let me know if I'm way off.

packages/web/src/shared/hooks/useLoadFonts.ts
Pertempto: Done

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -88,7 +89,10 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
                 value: gloss,
               }))}
               // The extra 24 pixel give room for the padding around the text.
-              style={{ width: width + 24, fontFamily: font }}
+              style={{
+                width: width + 24,
+                fontFamily: expandFontFamily(font ?? 'Noto Serif'),
```
arrocke: Let's fall back to Noto Sans instead.

arrocke (approved):
I think this should work well! Nice work

packages/web/src/features/translation/TranslateWord.tsx
Pertempto: So the default language font should also be Noto Sans?

arrocke: That's my thought. It's better for screens. We are using serif for Biblical texts, and sans serif for glosses and UI text

Pertempto:
@arrocke Don't forget to setup `NX_GOOGLE_FONT_API_KEY` before deploying

PR #207, opened 2023-09-25
feat: configure default reference translations for a language

### LINKED ISSUE #171: configure default reference translations for a language
A language admin should be able to set which translations from fetch.bible the language will use. It should provide them an autocomplete from the fetch.bible list to pick the translations. The user also needs to be informed how complete the translation is.

Pertempto:
## What has changed

Added a Bible translation selector to the manage language page.

## Connected Issues

Closes #171 

## QA Steps
1. Go to the manage language page for a language
2. Click on the bible translation dropdown
3. Select a few translations
4. Click the "submit" button
5. Refresh the browser page
    - The selects translations should persist

## Post-Deployment

- [ ] Nothing required

## Saved for Future Development
1. Allow user to re-order selected translations
    - #208
7. Allow user to filter translation options list
8. Display completion percentage for a given translation (I figured this would fit better with the UI for point 1)
    - #211


Pertempto:
## To Do
- [x] Actually get the list of translation options from fetch.bible

packages/web/src/shared/bibleTranslationClient.ts
```diff
@@ -0,0 +1,41 @@
...
+export type BibleTranslation = {
+  id: string;
+  name: string;
+};
+
+// For some reason, fetch.bible uses non-standard language codes.
+const langCodeMapping: Record<string, string> = {
+  en: 'eng',
+  es: 'spa',
+};
```
Pertempto, line 12: I don't know why `fetch.bible` seems to use English-based IDs, but we're going to have to add mappings for all the common languages, otherwise our users will have to scroll through 900+ translation options, which isn't a great experience. In the future it would be great to incorporate some kind of filtering logic to allow users to search the list of available translations.

package.json
```diff
@@ -11,6 +11,7 @@
     "@aws-sdk/client-sqs": "3.405.0",
     "@casl/ability": "6.5.0",
     "@casl/prisma": "1.4.0",
+    "@gracious.tech/fetch-client": "^0.5.3",
```
arrocke: Let's lock this version down

packages/web/src/shared/bibleTranslationClient.ts
arrocke: Looks like these are using [ISO 639-3](https://en.wikipedia.org/wiki/ISO_639-3) codes, which are more precise than the two letter ISO 639-1. I think for external integrations we are probably going to need to transpose between codes, but I wonder if ISO 639-3 is a better standard for us to use to identify a language. If we are going to do that, we should do so in a separate PR

packages/api/public/spec/schemas/language.yml
```diff
@@ -2,15 +2,15 @@ allOf:
   - $ref: './create-language.yml'
   - type: object
     properties:
-      code:
```
arrocke, line 5: Why did these get removed?

arrocke (commented):
I did a code review, but I haven't tested it yet.

packages/api/public/spec/schemas/language.yml
Pertempto: Because I moved them to `create-language.yml`. Now `language.yml` extends `create-language.yml`

packages/web/src/shared/bibleTranslationClient.ts
Pertempto: Ok, I'll just leave it as is for now, we can remove the mapping when we do that.

arrocke (approved):
I think the only thing this needs to is to set a max height on the dropdown so it can't scale past the end of the screen and then make it scrollable.

PR #213, opened 2023-10-03
feat: lemma resources route

### LINKED ISSUE #166: API routes to load lexicon data
* Add a new route `GET /lemma/:lemmaId/resources`
* This route is going to return all lexicon entries for a certain lemma
* Return data in the following shape:
```
{
  data: {
    resource: string
    entry: string
  }[]
}
```

Pertempto:
## What has changed

Added the lemma resources API route and documentation.

## Connected Issues

Closes #166 

## QA Steps

- [ ] Add a row or a few rows to the `LemmaResource` table
- [ ] Query for that data using the route `/api/lemmas/[lemmaId]/resources`

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/api/public/spec/schemas/parameters.yml
+++ b/packages/api/public/spec/schemas/parameters.yml
@@ -30,3 +30,11 @@ userId:
   schema:
     type: string
   example: '123e4567-e89b-12d3-a456-426614174000'
+lemmaId:
+  name: lemmaId
+  in: path
+  description: TODO - add description here.
+  required: true
+  schema:
+    type: string
+  example: G314
--- a/packages/api/public/spec/schemas/parameters.yml
+++ b/packages/api/public/spec/schemas/parameters.yml
@@ -33,7 +33,7 @@ userId:
 lemmaId:
   name: lemmaId
   in: path
-  description: TODO - add description here.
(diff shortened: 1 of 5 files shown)


packages/api/public/spec/schemas/parameters.yml
```diff
@@ -30,3 +30,11 @@ userId:
   schema:
     type: string
   example: '123e4567-e89b-12d3-a456-426614174000'
+lemmaId:
+  name: lemmaId
+  in: path
+  description: TODO - add description here.
```
Pertempto: I need a good description here. Any ideas?

arrocke:
This is in draft mode. Are you wanting a full review, or just general feedback?

Pertempto:
> This is in draft mode. Are you wanting a full review, or just general feedback?

I was asking for feedback. See the comment I left on the code

packages/api/public/spec/schemas/parameters.yml
arrocke: I'd just say something like `The id of the lemma as a strongs number`

Pertempto:
It should be ready now.

arrocke (approved):
Looks good

Issue #189, opened 2023-09-06
Automatic advance to next verse


BethedenMinistries:
When I save the last gloss in a verse it doesn't automatically take me to the next verse. 
So, if a user has already approved/saved all the glosses in a verse, and they hit Enter/Tab on the last word in the verse, it advances automatically to the next verse.
That's definitely an important feature to have before people start using it, since everyone is used to Sheldon's app doing that.

arrocke:
Totally agree on making that navigation easy. However, It's generally considered bad practice to hijack the tab sequence like that. That's the universal standard for navigating through interactive elements with just a keyboard.

What I would propose we do instead is add a button to the end of the verse line so that when you tab off of the last word you can just hit Enter on that button to navigate to the next verse. This would not disrupt the tab sequence allowing you to tab past the button to interactive elements further down the page.

BethedenMinistries:
Sounds good to me.

arrocke:
[Design](https://www.figma.com/file/GYKTJWl1otgwpmA9ZLLthE/Gloss-Translation-App?type=design&node-id=1218-701&mode=design&t=4tqE0ADsS4nGJ9XJ-0) for this is done. This is ready to be worked on.

## acceptance criteria
* At the end of the translation line add a link to navigate to the next verse
* The text should be vertically aligned with the text of the gloss line

<img width="606" alt="Screen Shot 2023-09-27 at 10 03 28 PM" src="https://github.com/arrocke/gloss-translation/assets/11968925/6ff7a467-8c73-4119-84d6-394222773983">

PR #215, opened 2023-10-04
refactor: cleaner gloss autocomplete input

### LINKED ISSUE #182: Use Enter key to save glosses
The Enter key should also save glosses (along with Tab). That avoids data loss, since many people are used to Enter already.

### LINKED ISSUE #183: Glosses not being accepted or reverting back
**Describe the bug**
When I attempted to replace a gloss of אֲשֶׁ֤ר in Ps 148.4 the gloss kept reverting back to what it originally was whenever I would press Tab. The gloss that I was trying to type in manually was already in the list, so I attempted to select it from the list, and it reverted again. When I selected it from the list a second time it finally staid.  

The dropdown menu for a gloss should not activate unless told to do so, since it interrupts the flow of the typing. When something is manually typed to replace a gloss, it should replace the gloss immediately without the drop down menu suggesting something automatically. This caused a lot of problems for me. 90% of the glosses I typed in to replace bad glosses were deleted by this function. 

I was using Chrome and the interface was in English and the glosses I was working on were the English ones.

arrocke:
## What has changed

* Current autocomplete input is renamed to be combobox input
* Autocomplete input is rebuilt to work more like a textbox with suggestions, similar to how search autocomplete works in most websites
* Additionally, the enter key will move between glosses. In future work, the enter key will be used for approval

## Connected Issues

closes #183, closes #182

## QA Steps

1. Verify that if you type something in the textbox, the tab and enter keys will always save that entry
2. Verify that an option is not highlighted until you use the arrow keys
3. Verify that when an option is highlighted, tab and enter select that value
4. Verify that typing filters the list and removes the item highlight

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -22,7 +22,6 @@ export interface TranslateWordProps {
   originalLanguage: 'hebrew' | 'greek';
   status: 'empty' | 'saving' | 'saved';
   onGlossChange(gloss?: string): void;
-  onKeyDown?: KeyboardEventHandler<HTMLElement>;
 }
 
 export interface TranslateWordRef {
@@ -40,7 +39,6 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       referenceGloss,
       previousGlosses,
       onGlossChange,
-      onKeyDown,
     }: TranslateWordProps,
     ref
   ) => {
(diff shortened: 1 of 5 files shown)


Pertempto:
Should an enter press on the last word of verse go to the next verse?

Pertempto (approved):
It seems to work well! I haven't been able to test the dropdown much, because my DB is empty of glosses. I'll need to get the import process working in my local env to test that.

arrocke:
> Should an enter press on the last word of verse go to the next verse?

That's going to be handled in as a separate PR for #189

PR #217, opened 2023-10-05
feat: validate language codes

### LINKED ISSUE #170: validate language codes
I wonder if it would be useful to provide a list of language codes that we accept to confirm they match the language code spec. We can use that list to pick fonts, default translations, etc automatically if we wanted to

Pertempto:
## What has changed

The new language page enforces using iso 639-3 language codes.

## Connected Issues

Closes #170 

## QA Steps

On the new language page, the code input should be an autocomplete. You should not be able to create a language with an invalid code.

## Post-Deployment

- [ ] Manually update production database to use valid iso 639-3 language codes for all the languages we have so far.


Pertempto:
@arrocke This isn't quite ready, because the `AutocompleteInput` isn't quite hooked up yet.

The new `AutocompleteInput` component doesn't seem to automatically work with forms. I could manually hook up the `onChange` and `value` props, but it seems better if we got the `AutocompleteInput` to play nice with forms automatically, just like all our other custom form components.

arrocke:
I'd switch to the ComboboxInput component instead since the AutocompleteInput is meant for suggestions, while ComboboxInput is closer to a select input but with filtering.

I'd look at what I did with the MultiselectInput and follow that as an example to modify the ComboboxInput to allow it to work with forms.

Pertempto:
I tried imitating `MultiselectInput`, but now I get the error 
```
A component is changing from uncontrolled to controlled. This may be caused by the value changing from undefined to a defined value, which should not happen.
```

Pertempto:
> This may be caused by the value changing from undefined to a defined value, which should not happen.

I got this working by providing `defaultValues` to `useForm`. Is there some better way to do this? We don't need it for the `MultiselectInput`

arrocke:
I pushed up a commit to address this without needing a default value. In react you generally have to work with inputs in one of two ways:

1. controlled - use a `value` prop and `onChange` event to track state outside the input
2. uncontrolled - use a `defaultValue` prop to set the initial value, but on submit read the data from the input element itself

If react notices that the `value` prop is undefined and you have an `onChange` event, then it throws that error because it thinks you aren't using the `value` prop with the `onChange` prop, so you are mixing these two methods.

This commit fixes that by ensuring that `value` is never undefined by defaulting to an empty string in that case

Pertempto:
Thanks for the explanation! Very helpful

Pertempto:
I don't see that your new commit, but otherwise this should be ready for review.

packages/web/src/shared/languageCodes.ts
```diff
@@ -0,0 +1,7920 @@
+// This list of iso639-3 codes is from SIL.
```
Pertempto, line 1: I seen something on SIL's site about requiring attribution, so hopefully this covers that.

packages/web/src/features/languages/NewLanguageView.tsx
```diff
@@ -92,7 +98,9 @@ export default function NewLanguageView() {
             />
           </div>
           <div>
-            <Button type="submit">{t('common:create')}</Button>
+            <Button type="submit" disabled={!formContext.formState.isValid}>
```
arrocke: Rather than disabling the submit button, I prefer to allow users to click the button. Our form setup prevents the submit from continuing while there are still validation errors and focuses the first element with errors so that it's easier to discover problems in the form, especially with screen readers.

packages/web/src/shared/components/form/ComboboxInput.tsx
```diff
@@ -115,22 +147,26 @@ const ComboboxInput = forwardRef<HTMLInputElement, ComboboxProps>(
...
-                    <span className="italic">{item.label}</span>"
-                  </>
-                ) : (
-                  item.label
-                )}
-              </Combobox.Option>
-            ))}
+            {filteredItems.length > MAX_ITEMS ? (
+              <div className="px-3 py-2">{t('common:too_many_options')}</div>
+            ) : (
```
arrocke, line 152: Rather than showing too many options, what do you think about just truncating to 100 items or so in a scrollable list? That way there are at least some options in the list

arrocke (approved):
Two minor suggestions, otherwise this is good

arrocke:
One other thing I thought of. Can we generate a new seed file with `eng` for the english code?

packages/web/src/shared/components/form/ComboboxInput.tsx
Pertempto: I don't like the idea of truncating, because of this particular type of user: They won't type in the input, and instead manually scroll the list looking for the option they want. I've encountered this at work, no matter how nice you make the filtering features, their first instinct is to manually scroll until they find the option they want.

Ideally, we would implement infinite scroll here, so that they could scroll until they found the option they needed. I didn't want to tackle that in this issue.

Pertempto: https://github.com/arrocke/gloss-translation/issues/222

Pertempto:
> One other thing I thought of. Can we generate a new seed file with `eng` for the english code?

That would be good, I don't have a good local database to generate from.

PR #218, opened 2023-10-06
feat: approve glosses

### LINKED ISSUE #203: Ability to make a gloss unconfirmed
Sometimes someone will accidentally or purposefully confirm a gloss as correct and then change their mind and want to leave it unconfirmed so that they can come back to it later after more research, etc. It would be nice to be able to use a shortcut key to unconfirm a gloss that has already been confirmed.

arrocke:
## What has changed

* Glosses now have a state with values `UNAPPROVED` and `APPROVED`
* Enter approves the gloss, escape unapproved
* Gloss suggestion dropdown is only previously approved glosses

## Connected Issues

closes #203 

## QA Steps

1. Confirm that enter approves the gloss
2. Confirm that tab does not approve
3. Confirm that selecting suggestion with mouse approves gloss
4. Confirm that escape unapproves gloss
5. Confirm that suggestion list is just previously approved glosses

## Post-Deployment

- [x] Nothing required


packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -116,10 +130,10 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
                       {capitalize(t('translate:saving'))}
                     </>
                   );
-                } else if (status === 'saved') {
+                } else if (status === 'approved') {
                   return (
                     <>
-                      <Icon icon="check" className="me-1" />
+                      <Icon icon="check" className="me-1 text-green-600" />
                       {capitalize(t('translate:saved'))}
```
arrocke: I forgot I need to update this string to say approved. That shouldn't take to long

packages/web/src/shared/components/form/AutocompleteInput.tsx
```diff
@@ -3,8 +3,17 @@ import { Icon } from '../Icon';
 
 export interface AutocompleteInputProps
   extends Omit<ComponentProps<'input'>, 'value' | 'onChange'> {
+  state?: 'success';
   value?: string;
-  onChange(value: string): void;
+  /** A change is implicit if it occurs:
```
Pertempto, line 8: Thanks for the clear comment! 👍🏽

arrocke:
Ok, this is ready now

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -286,16 +294,28 @@ export default function TranslationView() {
...
                     originalLanguage={isHebrew ? 'hebrew' : 'greek'}
                     status={
-                      isSaving ? 'saving' : targetGloss ? 'saved' : 'empty'
+                      isSaving
+                        ? 'saving'
+                        : targetGloss.gloss
+                        ? targetGloss.state === 'APPROVED'
+                          ? 'approved'
+                          : 'saved'
+                        : 'empty'
```
Pertempto: Nested ternaries get really hard to follow when you're trying to read them, could we move this logic to a helper function with `if` blocks?

packages/web/src/shared/components/form/AutocompleteInput.tsx
```diff
@@ -112,9 +134,9 @@ const AutocompleteInput = forwardRef<HTMLInputElement, AutocompleteInputProps>(
                   case 'Enter':
                   case 'Tab': {
                     if (typeof activeIndex === 'number') {
-                      change(filteredSuggestions[activeIndex]);
+                      change(filteredSuggestions[activeIndex], e.key === 'Tab');
```
Pertempto, line 148:
```suggestion
                      change(filteredSuggestions[activeIndex], true);
```

packages/web/src/shared/components/form/AutocompleteInput.tsx
```diff
@@ -112,9 +134,9 @@ const AutocompleteInput = forwardRef<HTMLInputElement, AutocompleteInputProps>(
...
                     } else {
-                      change(input);
+                      change(input, e.key === 'Tab');
```
Pertempto, line 150:
```suggestion
                      change(input, true);
```

Pertempto (changes requested):
A few minor things suggested above. I can't test until `main` is merged in, because the frontend is trying to use `en` for the default language code.

Pertempto:
Scratch that, I can test by changing my English language code back to `en` in my DB.

One bug I found:

1. Go to a verse full of empty glosses, one that you've never gone to before.
1. Repeatedly click the dropdown button on the the first verse


For some reason, the "saving" status shows up on another word of the verse. Each time you click the dropdown button on the first verse, the "saving" status moves to a different word in the verse. Very strange.

https://github.com/arrocke/gloss-translation/assets/17209828/28608aa9-e708-4deb-bce7-46e493fa6777


It only seems to happen the FIRST time I try it on a particular verse, so maybe it has something to do with the default state of words in the DB.

packages/web/src/shared/components/form/AutocompleteInput.tsx
arrocke: This condition is on purpose because we don't want the Tab key to signify approval since it used for navigation between interactive elements. In that sense, Tab is "implicit" because it is the keyboard version of clicking out of the input without making a selection

arrocke:
Ok, I think I got that bug fixed. This is ready for another review

packages/web/src/shared/components/form/AutocompleteInput.tsx
Pertempto: Yeah I understood that. This is inside a `case 'Tab':` block, so the expression `e.key === 'Tab'` will always evaluate to `true`. I should've made that clear in my original comment.

packages/web/src/features/translation/TranslationView.tsx
Pertempto: Much better.

packages/web/src/shared/components/form/AutocompleteInput.tsx
```diff
@@ -80,23 +81,25 @@ const AutocompleteInput = forwardRef<HTMLInputElement, AutocompleteInputProps>(
...
-          } else {
-            newValue = input;
-          }
-          if (newValue !== value) {
-            onChange(newValue, true);
+      if (isFocused) {
+        const handler = (e: PointerEvent) => {
+          if (!root.current?.contains(e.target as Element)) {
+            let newValue;
+            if (typeof activeIndex === 'number') {
```
Pertempto, line 88: I think our code would be much more readable if we defined a common helper function `isDefined`, and used it throughout our codebase.

```
export const isDefined = (value) => typeof value !== 'undefined';
```

Pertempto (changes requested):
It's looking pretty good. There are three remaining threads about code improvements, and I found another bug.

# Bug

1. Type a gloss for word
2. Press "Enter" to approve the gloss
3. Click on the word again
4. Enter a new word
5. Press "Tab" to go to the next word
    - The new gloss gets the "Approved" status, even though the user didn't indicate the new gloss should be approved.

Changing the gloss for a word doesn't automatically change the state back to "Unapproved". It seems like it should?

packages/web/src/shared/components/form/AutocompleteInput.tsx
arrocke: There are two cases for that block, both `Enter` and `Tab`. That's why there is that condition because it could be either key

packages/web/src/shared/components/form/AutocompleteInput.tsx
arrocke: If it was a more complex expression other than one comparison, I'd agree, but in this case, I don't think the cost of introducing a utility that you have to import is worthwhile.

packages/web/src/shared/components/form/AutocompleteInput.tsx
Pertempto: Ahhh, THAT is what I was missing. 🤦🏽  My bad.

Pertempto (approved):
It seems to work well. I found another bug, but I'm not sure that's it directly connected to these changes. I figured it was better to save it for a separate issue: https://github.com/arrocke/gloss-translation/issues/226

PR #219, opened 2023-10-07
feat: add next button in tab sequence at the end of the verse

### LINKED ISSUE #189: Automatic advance to next verse

arrocke:
## What has changed

* Added a button and the end of the verse to make it quick to navigate to the next verse after tabbing through each word

## Connected Issues

closes #189 

## QA Steps

Tab through a verse in Hebrew and Greek and then press enter when on the next button.

## Post-Deployment

- [x] Nothing required


arrocke:
I still have some work to do to ensure the next button is always aligned with the gloss autocomplete

Pertempto (commented):
When you use the next button, the keyboard focus on the new verse remains on the next button. I would expect it to go to the first word of the new verse, otherwise that's really going to interrupt the flow of glossing.

A few things I would change about the design:
 - There should be more space between the icon and the text.
     - Maybe this is just a personal preference? I feel like a lot of our icons are "squished" throughout the app.
 - It would feel much more natural to have the next button vertically aligned with the gloss inputs.
     - I could be wrong, but I'd predict that horizontal row of gloss inputs is where a translator's eyes are going to spend the most time.

arrocke:
Fixed the focus issue. I definitely agree that ideally, we would align the next link with the gloss input. The problem is that because the both the reference and target gloss fonts are configurable, the text baseline of the gloss input is variable. That's why I used a vertical line to separate them. At some point if we can know the text height of the different fonts we could do this, but for now, I was going to leave it like this.

Pertempto:
That makes sense. If we aligned from the bottom, you'd only have to calculate based on the height of the gloss input, regardless of the reference height.

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -331,6 +344,31 @@ export default function TranslationView() {
                   />
                 );
               })}
+              {canEdit && (
+                <li
+                  className={`
+                    mx-2 mb-4 flex items-center border-slate-300 h-28
```
Pertempto:
```suggestion
                    mx-2 mb-4 flex flex-col border-slate-300 h-28
```

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -331,6 +344,31 @@ export default function TranslationView() {
...
+                    ${isHebrew ? 'border-r pr-4' : 'border-l pl-4'}
+                  `}
+                >
+                  <Button
+                    variant="tertiary"
+                    className="mb-2"
```
Pertempto:
```suggestion
                    className="mb-2 mt-auto"
```

Pertempto (approved):
The functionality works great. I added some suggestions that will align the next button to the bottom. It doesn't perfectly align with the gloss inputs, but I think it looks much more natural than aligning it vertically centered.

arrocke:
I did some more digging and I think the line height issue was just in the select input in the language page, so I think for now we can just set a fixed top margin for alignment. We can come back to this later if we need to

Pertempto (approved):
That looks good! 

Just curious, what is the advantage of a bunch of top margin, as compared to just aligning to the bottom?

arrocke:
You need spacing from the top or bottom either way. Aligning from the bottom requires an extra css property to set that alignment, and it also changes when the saving/approved indicator appears

PR #220, opened 2023-10-08
refactor: align gloss line according to the ancient language

### LINKED ISSUE #201: Gloss alignment
I hate to be nitpicky, but it would be nice to have the glosses align right underneath the Hebrew when you're in editing mode. The arrow menu dropdowns offset the glosses significantly, and so you're eyes have to keep looking back at an angle to see what you're glossing. My ideal would be to get rid of those arrows or find a way to align them better.

arrocke:
## What has changed

Align translation line to the left for greek and to the right for hebrew

## Connected Issues

closes #201 

## QA Steps

1. Check alignment for hebrew when the UI is in english and arabic. The translation line shouldn't change
2. Do the same for greek

## Post-Deployment

- [x] Nothing required


Pertempto:
It doesn't work correctly when the gloss is shorter than the English gloss. (notice "of the fathers" and "of the sons")

image

Funny enough, if I change the interface language to Arabic, they align correctly. You can see the inputs for those two glosses shift left/right when you change the interface language.

image


Overall, we've introduced some issues with gloss input widths sometime recently:

image

arrocke:
Ah, I see what is going on. There is a few issues with dealing with rtl vs ltr language that we need to handle. I create a separate issue (#223) to configure the language direction. We'll need to use that to configure how we render the gloss selector

Pertempto (approved):

PR #221, opened 2023-10-08
refactor: manually check email status before sending email changed message

### LINKED ISSUE #141: Separate email verification from users
In #140 we are sending emails to addresses that aren't associated with a user. This happens because we want to verify an email address before we attach it to a user, or because we want to send a notification to the previous email when it is changed by a user. This means we should separate users and verified emails in the database.

1. A new database table called `EmailVerification` with two columns `email` and `status`. `status` can be `VERIFIED`, `BOUNCED`, or `COMPLAINED`. This would replace the email verification column on the `AuthUser` table.
2. When we send an email we can look up the address against this table and only send if the status is verified.

arrocke:
## What has changed

I took another look at this issue and turns out the only time we need to confirm that an email address can receive email when sending to an address not associated with a user is after the user has changed their email address. In this case we are sending an email to the old address to notify them of the change. So I just modified that particular instance to verify the email address before we change it. In all other cases, we can look up the email verification status on the user before sending an email, so I'm not going to mess with this.

I also noticed two other issues that I fixed:
* If the user's email address had previously bounced, changing their address wasn't resetting their email status to verified so they could receive email again
* The verification token was being deleted once it was verified.

## Connected Issues

closes #141 

## QA Steps

1. Set your email status to bounced manually in the database
2. Change your email address in your profile
3. Click the link in the email
4. Verify that your email status is now verified, the verification token is removed, and the new email address is sent
5. Verify that the email change notification was sent

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/api/pages/api/email/verify.ts
+++ b/packages/api/pages/api/email/verify.ts
@@ -2,9 +2,9 @@ import * as z from 'zod';
 import createRoute from '../../../shared/Route';
 import { auth } from '../../../shared/auth';
 import { InvalidTokenError } from '../../../shared/errors';
-import { client } from '../../../shared/db';
+import { PrismaTypes, client } from '../../../shared/db';
 import { PostEmailVerificationRequest } from '@translation/api-types';
-import mailer from '../../../shared/mailer';
+import mailer, { EmailNotVerifiedError } from '../../../shared/mailer';
 
 export default createRoute()
   .post<PostEmailVerificationRequest, void>({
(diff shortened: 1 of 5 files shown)


Pertempto:
Does this break new user logic? I currently only have the default `devyn` account, so I tried to add another one.

I got this error in the backend:

```
EmailNotVerifiedError: We can't send emails to test@test.com because it is not verified.
    at Object.sendEmail (webpack-internal:///(api)/./shared/mailer.ts:58:111)
    at async Object.handler (webpack-internal:///(api)/./pages/api/users/index.ts:85:9)
    at async eval (webpack-internal:///(api)/./shared/Route.ts:152:21)
    at async eval (webpack-internal:///(api)/./shared/Route.ts:211:21)
```

arrocke:
Nice catch! That should be fixed now

Pertempto (approved):
This seems to work well!

Issue #83, opened 2023-05-12
vagrant setup for easy development


arrocke:
## acceptance criteria
* setup a vagrantfile to spin up a development environment that has the right node environment, and a postgres database with seed data already set up
* document this in contributing documentation

ilearner777:
I put the following request into chatGPT:
```
How do you setup a vagrant file to a create a windows 11 development environment that will perform the following tasks:
1. Install Node.js 18
2. In a terminal run: npm install
3. In a terminal run: npm i -g nx
4. Install postgresql version 14 (latest version).
5. Create database for the project. 
6. Create new file: ./packages/api/.env.local
7. Create new file: ./packages/db/.env.local
8. Add environment variable: DATABASE_URL to both files above to the postgres connection string. Here is a template you can use:
DATABASE_URL=postgres://{user}:{password}@{hostname}:{port}/{database-name}
9.  In a terminal run: nx prisma db migrate reset
10. In a terminal run: pg_restore -Fc --format=custom --dbname=<connection-string> data/seed.dump
```

This was response:

To set up a Vagrantfile for a Windows 11 development environment with the specified tasks, you can follow these steps:
```
    Create a directory for your Vagrant project and navigate to it in your terminal.

    Initialize a new Vagrant environment:

bash

vagrant init

    Edit the Vagrantfile to configure your Windows 11 environment:

ruby

Vagrant.configure("2") do |config|
  config.vm.box = "your-windows11-box"  # Replace with the Windows 11 box name or URL

  # Forward ports if needed
  config.vm.network "forwarded_port", guest: 80, host: 8080  # Example port forwarding

  # Set up provisioning script
  config.vm.provision "shell", inline: <<-SHELL
    # Install Node.js 18
    choco install nodejs-lts

    # Navigate to the project directory
    cd /path/to/your/project

    # Install npm packages
    npm install
    npm install -g nx

    # Install PostgreSQL 14
    choco install postgresql

    # Create database for the project
    psql -U postgres -c "CREATE DATABASE your-database-name"

    # Create .env.local files and add DATABASE_URL
    echo "DATABASE_URL=postgres://user:password@hostname:port/your-database-name" > ./packages/api/.env.local
    echo "DATABASE_URL=postgres://user:password@hostname:port/your-database-name" > ./packages/db/.env.local

    # Run database migrations
    nx prisma db migrate reset

    # Restore data from seed.dump (replace <connection-string>)
[...]

PR #225, opened 2023-10-10
chore: add vagrant support

ilearner777:
## What has changed

- Setup a Vagrantfile to spin up a development environment that has the right node environment, and a postgres database with seed data already set up.
- Documented this in the contributing documentation (with link to new vagrant.md)

## Connected Issues

#83  

## QA Steps
Follow instructions in the `Post-Installation Notes` section of file `vagrant.md`.
1. Make sure you can access database using pgAdmin.
2. Make sure you can ssh to VM.
3. Start API server in VM and make sure you can access via host web browser.
4. Start Web app in VM and make sure you can access via host web browser.

## Post-Deployment

- [X] Nothing required

## Known Issue

When accessing Web app via external browser, gets a runtime error when app attempts to get the English gloss for Gen. 1:1.


docs/vagrant.md
```diff
@@ -0,0 +1,104 @@
+# Automated Ubuntu Local Env Setup
+
+You can optionally setup a virtual Ubuntu environment using Vagrant. It is a tool used for building and managing virtual machine environments in a single workflow.
+
+Although Vagrant supports [VirtualBox](https://www.virtualbox.org/), [Hyper-V](https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/enable-hyper-v), [Docker](https://www.docker.com/), or [VMware Fusion](https://customerconnect.vmware.com/downloads/get-download?downloadGroup=FUS-PUBTP-2021H1), `VirtualBox` is the only environment that has been developed and tested thus far. If you have need to use Hyper-V, Docker, or VMware, please create an issue and request a new feature.
+
+## Prerequisites
+
+1. You will require 4 GB of memory.
+1. You will require 8 GB of diskspace.
```
Pertempto:
```suggestion
1. You will require 8 GB of disk space.
```

docs/vagrant.md
```diff
@@ -0,0 +1,104 @@
...
+1. Reference these [installation](https://developer.hashicorp.com/vagrant/docs/installation) instructions.
+1. Verify installation by running `vagrant -v` in a terminal session.
+1. Install [VirtualBox](https://www.virtualbox.org/).
+1. If you have not already done so, fork repo to your local machine.
+1. Create a new folder to contain your `Vagrant project`. Start a terminal session in that folder.
+1. Install a [box](https://developer.hashicorp.com/vagrant/tutorials/getting-started/getting-started-boxes). The `Vagrantfile` in this repository uses `generic/ubuntu2204`. To install it run:
+
+```bash
+vagrant box add generic/ubuntu2204
+```
```
Pertempto:
```suggestion
    ```bash
    vagrant box add generic/ubuntu2204
    ```
```
Tip: add indentation of four spaces before your code blocks to make them show up inline with your numbered list. (I've only commented on the one block of code, but this suggestion should be applied to all the code blocks within this numbered list)

docs/vagrant.md
```diff
@@ -0,0 +1,104 @@
...
+
+8. Copy file: `Vagrantfile` from host `gloss-translation` repository `.vagrant/` folder to your vagrant project folder.
+9. Read the `NOTES:` section contained in file `Vagrantfile`.
+10. in the `Vagrantfile`, edit `HOST_FOLDER`
+
+```ruby
+ HOST_FOLDER = "../gloss-translation"
+```
+
+Replace `../gloss-translation` with the location of your repository root folder.
```
Pertempto:
```suggestion
    Replace `../gloss-translation` with the location of your repository root folder.
```

docs/vagrant.md
```diff
@@ -0,0 +1,104 @@
...
+   1. Symbolic links should now work in VirtualBox shared folders without administrator priviledges.
+
+## Setup Steps
+
+1. [Download](https://developer.hashicorp.com/vagrant/downloads?ajs_aid=ba208f95-d5a1-457d-abeb-49d458b95eec&product_intent=vagrant) Vagrant software.
+1. Reference these [installation](https://developer.hashicorp.com/vagrant/docs/installation) instructions.
+1. Verify installation by running `vagrant -v` in a terminal session.
+1. Install [VirtualBox](https://www.virtualbox.org/).
+1. If you have not already done so, fork repo to your local machine.
+1. Create a new folder to contain your `Vagrant project`. Start a terminal session in that folder.
```
Pertempto:
```suggestion
1. Create a new folder to contain your Vagrant project. Start a terminal session in that folder.
```
It's not clear: should my vagrant project folder be inside my repo? Or somewhere else?

Pertempto (changes requested):
I've installed vagrant and virtualbox, and have worked my way to step 6. These are my first suggestions, and I await your response to [my question](https://github.com/arrocke/gloss-translation/pull/225#discussion_r1356457967).

docs/vagrant.md
ilearner777: Done.

ilearner777: Done.

docs/vagrant.md
ilearner777: `Replace...` line is now integrated into the line before the code block.

docs/vagrant.md
ilearner777: I have completely rewritten this section, improving clarity and providing details as to the purpose of the new folder to be created.

docs/vagrant.md
ilearner777: Done

arrocke:
I'm going to try and look at this in the next few days

docs/vagrant.md
Pertempto: You need to actually indent the whole block, not just the code inside the block. Here's an example of what it should look like:

image

Pertempto: This applies to all the code blocks in this file. Do you have a markdown preview in your VS code? If you preview the markdown, it will make it obvious where your indentation needs adjusted.

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+    vagrant box add generic/ubuntu2204
+```
+
+To see which boxes are installed run:
+
+```bash
+    vagrant box list
+```
+
+7. Create a new folder and name it something that identifies the VM you are about to create. Example: `UbuntuVM`. The name of this folder will be used by Vagrant to identify this specific VM. It is to this directory that you will copy the vagrant file we have created for you. In a later step, Vagrant will create subdirectory`.vagrant` which it will use to manage your VM. It is in this directory where you will run all Vagrant commands that are specific to your VM.
```
Pertempto: @arrocke It almost sounds to me like we should check `UbuntuVM` (or another name) into the git repo, and not bother with having a top-level `.vagrant` directory that people have to copy from.

What do you think?


@ilearner777 Would my idea cause any issues?

docs/vagrant.md
```diff
@@ -23,45 +23,43 @@ Although Vagrant supports [VirtualBox](https://www.virtualbox.org/), [Hyper-V](h
 1. Verify installation by running `vagrant -v` in a terminal session.
 1. Install [VirtualBox](https://www.virtualbox.org/).
 1. If you have not already done so, fork repo to your local machine.
-1. Create a new folder to contain your `Vagrant project`. Start a terminal session in that folder.
-1. Install a [box](https://developer.hashicorp.com/vagrant/tutorials/getting-started/getting-started-boxes). The `Vagrantfile` in this repository uses `generic/ubuntu2204`. To install it run:
+1. Next you will need to install a [box](https://developer.hashicorp.com/vagrant/tutorials/getting-started/getting-started-boxes). The `Vagrantfile` in this repository uses box: `generic/ubuntu2204`. So to install this box, run in a terminal window:
 
 ```bash
-vagrant box add generic/ubuntu2204
+    vagrant box add generic/ubuntu2204
```
Pertempto: You mentioned virtualbox before, but it would be good to mention it again here so people don't get confused when vagrant asks them which provider to use.

docs/vagrant.md
Pertempto: I think if we did this, we could remove point 10 and just have `HOST_FOLDER = "../"` in the `Vagrantfile`

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+### Accessing Web app
+
+Open a web brower to URL: <http://localhost:4200>
+
+### Access VM via ssh
+
+- host Name: `localhost`
+- Port: `2222`
+- login name: `vagrant`
+- password: `vagrant`
```
Pertempto:
```suggestion
| Setting  | Value  |
| ------------- | ------------- |
|  hostname | `localhost`  |
| port  | 2222  |
| username  | `vagrant`  |
| passowrd  | `vagrant`  |
```

Pertempto: You could also use table formatting for the Connection tab item and SSH Tunnel tab settings, if you want.

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+1. `General` tab:
+   - Name: _Enter name of your choice_
+1. `Connection` tab
+
+   - Host name: `127.0.0.1`
+   - Port: `5432`
+   - Maintenance database: `postgres`
+   - Username: `vagrant`
+
+1. `SSH Tunnel` tab
```
Pertempto: You forgot to mention the value for "Tunnel host".

Also, could we save the password in this tab to save time later?

Pertempto (changes requested):
Things seem to be working well, my VM is starting up now. A few improvement suggestions.

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+    HOST_FOLDER = "../gloss-translation"
+```
+
+11. If required, perform any requested `plugin` installations mentioned in the `NOTES:` section. To see which plugins are installed run:
+
+```bash
+    vagrant plugin list
+```
+
+12. To start the creation of your virtual development environment, run this command in your newly created directory:
```
Pertempto: It's not very clear, will we also run this to start the VM in the future, or is this command just for the initial creation of the virtual machine? How do we safely stop/shutdown the VM?

docs/vagrant.md
ilearner777: I see there is a disconnect in our conversation.  This has to do with how I setup my directory structure when learning and developing this code with respect to what you are visualizing.  Let me explain...

In my directory structure, UbuntuVM is in a parallel directory to the repo.  This is why I have `HOST_FOLDER = .../"gloss-translation"` as the host URL. 
image
I believe you are visualizing the UbuntuVM folder within the gloss-translation repository. 
Is this correct?

If so, I believe this is a better approach and found out that this is the preferred method.

So I am suggesting the following changes to this PR:

***Revised Directory Structure***
image

***Notes:***
- I have added a sub-directory `virualbox` to folder `.vagrant`. This is where I have placed the Vagrantfile configured for a Virtualbox VM. This is because I have already developed a Vagrantfile for the Docker environment. I plan to add a new issue/PR for this new environment. At that time I plan to add another sub-directory under `.vagrant` called `docker` into which I will place the Vagrantfile with configuration to create a docker VM. 
image
- I have renamed directory `UbuntuVM` to `vm` in order to keep it more generic, just in case someone wants an O/S other than Ubuntu.
- I have updated the .gitignore file to exclude the contents of the `vm` directory.  This is because when the user runs `Vagrant up`, files will be created that are specific to that user's machine.  These should not be stored in the upstream repository. See excerpt below:

    ```text
    # Virtual Machine
    vm/*
    !vm/.gitkeep
    ```

Please let me know if you approve of this recommendation and if you would like to modify anything I have described above.  I will await your reply before proceeding any further.

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+   1. Symbolic links should now work in VirtualBox shared folders without administrator priviledges.
+
+## Setup Steps
+
+1. [Download](https://developer.hashicorp.com/vagrant/downloads?ajs_aid=ba208f95-d5a1-457d-abeb-49d458b95eec&product_intent=vagrant) Vagrant software.
+1. Reference these [installation](https://developer.hashicorp.com/vagrant/docs/installation) instructions.
+1. Verify installation by running `vagrant -v` in a terminal session.
+1. Install [VirtualBox](https://www.virtualbox.org/).
+1. If you have not already done so, fork repo to your local machine.
+1. Next you will need to install a [box](https://developer.hashicorp.com/vagrant/tutorials/getting-started/getting-started-boxes). The `Vagrantfile` in this repository uses box: `generic/ubuntu2204`. So to install this box, run in a terminal window:
```
arrocke: If vagrant has already been initialized, then you don't need to install the box. Running `vagrant up` will install the box if it isn't already installed.

docs/vagrant.md
arrocke: Personally, I think the Vagrantfile should be in the root directory and run from the root directory. As soon as we have to copy it, now there are going to be syncing issues if users haven't copied the latest version. We can add `.vagrant/` to `.gitignore` since that is produced at runtime.

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+7. Create a new folder and name it something that identifies the VM you are about to create. Example: `UbuntuVM`. The name of this folder will be used by Vagrant to identify this specific VM. It is to this directory that you will copy the vagrant file we have created for you. In a later step, Vagrant will create subdirectory`.vagrant` which it will use to manage your VM. It is in this directory where you will run all Vagrant commands that are specific to your VM.
+8. Copy file: `Vagrantfile` from host `gloss-translation` repository `.vagrant/` folder to the new folder you created above.
+9. Read the `NOTES:` section contained in file `Vagrantfile`.
+10. in the `Vagrantfile`, edit `HOST_FOLDER` by replacing `../gloss-translation` with the location of your repository root folder.
+
+```ruby
+    HOST_FOLDER = "../gloss-translation"
+```
+
+11. If required, perform any requested `plugin` installations mentioned in the `NOTES:` section. To see which plugins are installed run:
```
arrocke: The plugins that need to be installed should be documented here, not in the vagrant file

docs/vagrant.md
```diff
@@ -0,0 +1,102 @@
...
+    vagrant plugin list
+```
+
+12. To start the creation of your virtual development environment, run this command in your newly created directory:
+
+```bash
+    vagrant up
+```
+
+13. For help on vagrant commands, at a terminal session enter:
```
arrocke: We should mention that you need to run `vagrant ssh` to get access to the build environment to start the dev servers.

.vagrant/Vagrantfile
```diff
@@ -0,0 +1,196 @@
...
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://vagrantcloud.com/search.
+  config.vm.box = "generic/ubuntu2204"
+
+  # Configure for application API
+  config.vm.network "forwarded_port", guest: 4300, host: 4300, host_ip: "127.0.0.1"
+
+  # Configure for application Web
+  config.vm.network "forwarded_port", guest: 4200, host: 4200, host_ip: "127.0.0.1"
```
arrocke: Can we also forward the postgres port so that we don't need to use an SSH tunnel for pgadmin?

.vagrant/Vagrantfile
```diff
@@ -0,0 +1,196 @@
...
+
+  # Mount repo project root folder.
+  config.vm.synced_folder HOST_FOLDER, "/home/vagrant/vmrepo", type: "rsync",
+    rsync__exclude: [".git/",".vagrant/","docs/", "node_modules/","packages/api/.env.local","packages/db/.env.local"]
```
arrocke: We will want to sync over the .env.local files so that the user can modify env vars. We can set the env vars we need to in this file, but we need to be able to set other secrets that we don't want to check into the public repo

.vagrant/Vagrantfile
```diff
@@ -0,0 +1,196 @@
...
+  config.vm.provision "shell", name: "npm", privileged: false, inline: <<-SHELL
+    echo "Setting up npm"
+    echo "Provisioning with user access"
+    echo "cd /home/vagrant/vmrepo"
+    cd /home/vagrant/vmrepo
+
+    echo "Performing npm install..."
+    npm install
+    echo "Setting nx as global..."
+    sudo npm i -g nx
```
arrocke: We need to install nx first so that when `npm install` is run the `nx` command is available to execute the `postinstall` script in package.json. This is what is causing the 500 error you were seeing.

.vagrant/Vagrantfile
```diff
@@ -0,0 +1,196 @@
...
+    echo "npm provisioning completed."
+
+    echo "Recompiling project..."
+    nx build api
+    nx build web
```
arrocke: We don't need to compile the project because the developer will use the build scripts to execute so that they get hot reload when files change.

arrocke (changes requested):
Thanks for your work on this. Getting this set up is quite complicated. Here are a few of my thoughts after looking through the PR.

docs/vagrant.md
ilearner777: Your recommendation to put Vagrantfile in the root directory is fine with me and I see the benefit of it.  However, I also see a downside in that then, we could only support a single provider. This is because we could only support a single version of the Vagrantfile.  

I have already created and successfully tested another version of the Vagrantfile that creates a VM using docker.  If you want to offer contributors the option to choose between multiple providers we would have to come up with a different approach. This was the reason why my design had the contributor copy the Vagrantfile of their choice to the VM folder.

I am good with whatever you wish to do.  Please let me know if you want to stick with offering only VirtualBox, or reconfigure with the  option to choose between multiple providers.

arrocke: [This from the vagrant ](https://developer.hashicorp.com/vagrant/docs/providers/configuration)docs suggests that multiple providers is not an issue. These lines from the vagrantfile are ignored if you use a different provider:
```ruby
config.vm.provider "virtualbox" do |vb|
    vb.memory = 4096  # Set RAM in MB (4GB in this example)
    vb.cpus = 2      # Set the number of CPU cores
    vb.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate//home/vagrant/vmrepo", "1"]
  end
```

Everything else should be the same regardless of provider and anything that is required for only VirtualBox should be within that block

ilearner777: Thanks for pointing that out to me.  Was not aware of this.  So I will update Vagrantfile so it is compatible with both VirtualBox and Docker. Then will implement all of your recommendations.  When complete will request another review.

.vagrant/Vagrantfile
ilearner777: There is a problem if I remove `"packages/api/.env.local","packages/db/.env.local"` from the exclude list.
Namely, plugin `vagrant-gatling-rsync` is only a one-way sync from host to guest.  During the initial run of `vagrant up`, both .env.local files are created in guest environment only as there is no replication from guest to host.  When the inital `vagrant up` install is completed, Vagrant autostarts the  `vagrant-gatling-rsync` from host to guest. Consequently it deletes the 2 new .env.local files because they do not exist on the host side.

We can make the change you requested, however, as a work-around to this problem, we would have to update the user instructions to have them run the following 2 steps after the initial install shown below:  
The first line recreates both .env.local files. 
 The 2nd line syncs from guest to host.

```
vagrant provision --provision-with env
vagrant rsync-back
```

.vagrant/Vagrantfile
ilearner777: I added the following line:

```ruby
config.vm.network "forwarded_port", guest: 5432, host: 5432, host_ip: "127.0.0.1"
```
however, when I tested using PgAdmin, I got the following error;

```text
connection failed: :1), 
port 5432 failed: could not receive data from server: Socket is not connected (0x00002749/10057) could not send SSL negotiation packet: Socket is not connected (0x00002749/10057)
```

I was able to resolve this problem by modifying the following postgres files: `postgresql.conf` and `pg_hba.conf`.  
To implement this changes, I would need to add this logic to the Vagrantfile dbsetup provisioning section.

2 Questions:
1. Do you want me to implement this and remove SSH Tunnelling instructions?
2. Do you want me to keep port at default 5432, or change to something else (5435?) just in case user has postgres running on host?

arrocke: Let's modify the config files and remove the ssh tunneling

.vagrant/Vagrantfile
arrocke: Can you explain more the need for rsync? Doesn't vagrant sync folders out of the box without a plugin? https://developer.hashicorp.com/vagrant/docs/synced-folders

docs/vagrant.md
ilearner777: I believe this only occurs when you use command `vagrant init`.  When I retested  on 2023-10-28 following my latest instructions, vagrant did not ask me which provider to use.

docs/vagrant.md
ilearner777: I removed the install box section as you suggested.

docs/vagrant.md
ilearner777: Done.

.vagrant/Vagrantfile
ilearner777: Done.

.vagrant/Vagrantfile
ilearner777: Originally, I tried with default virtualbox syncing, but had issues with symbolic links. So this is why I switched to rsync. However, I retried default virtualbox at your request, and now it works fine.  Default is simpler and faster.  Also, default virtualbox works differently than rsync.  rsync does replication, where virutalbox performs a mount. As a result, you cannot exclude any directories with virtualbox.

.vagrant/Vagrantfile
ilearner777: Done.

.vagrant/Vagrantfile
ilearner777: Removed those commands.

docs/vagrant.md
```diff
@@ -0,0 +1,134 @@
...
+   vagrant up
+   ```
+
+1. Once the `vagrant up` command completes, you can logon into your virutal machine using this command. (password: `vagrant`)
+
+   ```bash
+   vagrant ssh
+   ```
+
+1. Here is a list of useful Vagrant commands. Be sure that you are in the same directory as the Vagran­tfile when running these commands!
```
Pertempto: Is it necessary to have this list, when we can just use `vagrant --help` to get the available commands? Having this list in our docs requires us to keep it in sync with vagrant's docs. If they update or remove a command, our docs will be out of date.

docs/vagrant.md
```diff
@@ -0,0 +1,134 @@
...
+   | `vagrant box remove` [virtualbox name]        | delete a box                                                                                                                         |
+   | `vagrant global­-status`                      | outputs status of all vagrant machines                                                                                               |
+   | `vagrant global­-status` --prune              | same as above, but prunes invalid entries                                                                                            |
+   | `vagrant up` --provision \| tee provision.log | Runs vagrant up, forces provisioning and logs all output to a file                                                                   |
+   | `VAGRANT_LOG=info`                            | Use the environement variable VAGRANT_LOG to set verbosity. Enter this in terminal session before you run the `vagrant up` command   |
+
+1. For help on vagrant commands, at a terminal session enter:
+
+   ```bash
+   vagrant -help
```
Pertempto:
```suggestion
   vagrant --help
```

docs/vagrant.md
Pertempto: Some of the items like `VAGRANT_LOG=info` would be useful for our docs, because they are not listed in `vagrant --help`

docs/vagrant.md
```diff
@@ -0,0 +1,134 @@
...
+   1. Symbolic links should now work in VirtualBox shared folders without administrator priviledges.
+
+## Setup Steps
+
+1. [Download](https://developer.hashicorp.com/vagrant/downloads?ajs_aid=ba208f95-d5a1-457d-abeb-49d458b95eec&product_intent=vagrant) Vagrant software.
+1. Reference these [installation](https://developer.hashicorp.com/vagrant/docs/installation) instructions.
+1. Verify installation by running `vagrant -v` in a terminal session.
+1. Install [VirtualBox](https://www.virtualbox.org/).
+1. If you have not already done so, fork repo to your local machine.
+1. Vagrant uses a special configuration file called: `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host) . It is located n your repo root directory.
```
Pertempto:
```suggestion
1. Vagrant uses a special configuration file called `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host). It is located in the root directory of the repo.
```

docs/vagrant.md
```diff
@@ -0,0 +1,134 @@
...
+   | `vagrant up` --provision \| tee provision.log | Runs vagrant up, forces provisioning and logs all output to a file                                                                   |
+   | `VAGRANT_LOG=info`                            | Use the environement variable VAGRANT_LOG to set verbosity. Enter this in terminal session before you run the `vagrant up` command   |
+
+1. For help on vagrant commands, at a terminal session enter:
+
+   ```bash
+   vagrant -help
+   ```
+
+1. Once your VM environment is created, as an alternative to the vagrant commands listed above, you can use the `Oracle VM VirtualBox` application to manage the starting, stopping, and configuration of your VM. The `Help` menu provides detailed instructions on its use.
```
Pertempto: Will using Virtualbox directly break the vagrant setup?

It seems unnecessary to mention this, when we can just use vagrant to manage our VMs in a consistent manner, regardless of provider.

Pertempto (changes requested):

docs/vagrant.md
Pertempto: Maybe we should include the very basic commands like `vagrant up`, `vagrant halt`, and `vagrant ssh` (basically only the most basic commands which we'll need for day-to-day usage). Then we can refer developers to `vagrant --help` if they want to go deeper.

@arrocke What are your thoughts? Am I way off?

arrocke: Yep, this documentation should be enough for a developer to get started, and refer to the vagrant docs for more detail. I think we can assume that devs that want to use vagrant, will do their own research or will already be familiar

docs/vagrant.md
arrocke: Opening virtualbox shouldn't break vagrant, sometimes that is useful for debugging what is going on. But I agree, this shouldn't be necessary in our docs

docs/vagrant.md
ilearner777: Done.

docs/vagrant.md
ilearner777: So just to be clear. Do you want me to remove point 9.  Here is a list of usefule Vagrant commands....    but leave point 10?

docs/vagrant.md
ilearner777: I remove point 11.

docs/vagrant.md
ilearner777: Done.

docs/vagrant.md
arrocke: I'm ok removing both. Devs can do their own research if they want to do more than the basic commands to get everything running

arrocke (approved):
Thanks for taking care of this! This is good to merge once @Pertempto approves

docs/vagrant.md
Pertempto: @ilearner777 Go ahead and remove points 9 and 10. You should replace them with a point that describes the basic commands, just "enough for a developer to get started" like Adrian said.

docs/vagrant.md
```diff
@@ -0,0 +1,132 @@
...
+   | Host name:            | `127.0.0.1` |
+   | Port:                 | `5432`      |
+   | Maintenance database: | `postgres`  |
+   | Username:             | `vagrant`   |
+   | Password:             | `vagrant`   |
+   | Save password?:       | `Turn on`   |
+
+### Accessing API app
+
+Open a web brower to URL: <http://localhost:4300>
```
Pertempto:
```suggestion
Open a web browser to URL: <http://localhost:4300>
```

docs/vagrant.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Accessing Web app
+
+Open a web brower to URL: <http://localhost:4200>
```
Pertempto:
```suggestion
Open a web browser to URL: <http://localhost:4200>
```

docs/vagrant.md
```diff
@@ -0,0 +1,132 @@
...
+1. Install [VirtualBox](https://www.virtualbox.org/).
+1. If you have not already done so, fork repo to your local machine.
+1. Vagrant uses a special configuration file called `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host) . It is located n your repo root directory.
+1. To start the creation of your virtual environment, run this command in the repo root directory. (This takes a while so be patient.)
+
+   ```bash
+   vagrant up
+   ```
+
+1. Once the `vagrant up` command completes, you can logon into your virutal machine using this command. (password: `vagrant`)
```
Pertempto:
```suggestion
1. Once the `vagrant up` command completes, you can logon into your virtual machine using this command. (password: `vagrant`)
```

I would recommend installing the "Code Spell Checker" extension in VS Code, so it is easier to catch these issues.

docs/vagrant.md
```diff
@@ -0,0 +1,132 @@
...
+   1. Symbolic links should now work in VirtualBox shared folders without administrator priviledges.
+
+## Setup Steps
+
+1. [Download](https://developer.hashicorp.com/vagrant/downloads?ajs_aid=ba208f95-d5a1-457d-abeb-49d458b95eec&product_intent=vagrant) Vagrant software.
+1. Reference these [installation](https://developer.hashicorp.com/vagrant/docs/installation) instructions.
+1. Verify installation by running `vagrant -v` in a terminal session.
+1. Install [VirtualBox](https://www.virtualbox.org/).
+1. If you have not already done so, fork repo to your local machine.
+1. Vagrant uses a special configuration file called `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host) . It is located n your repo root directory.
```
Pertempto:
```suggestion
1. Vagrant uses a special configuration file called `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host). It is located in your repo root directory.
```

You missed a few things in the last correction.

Pertempto: You can use the "Commit suggestion" button to avoid missing anything.

.vagrant/Vagrantfile
Pertempto: Could we include an item in the "Caveat" section that mentions that postgres running on the host needs to be stopped before starting the VM? I had postgres running on my host, and got this error
```
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to 5432 is already in use
on the host machine.
```
I stopped the postgres Windows service, and then the virtual machine ran fine. Maybe we should include a note about this in the docs, in case a future dev runs into the same error message.

Pertempto:
How long does it usually take to boot the VM? This is at the bottom of the output of `vagrant up`

```
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
```

I tried `vagrant ssh`, but it timed out. I tried to connect to the DB, but it timed out. I tried `vagrant halt`, but got this error:
```
An action 'halt' was attempted on the machine 'default',
but another process is already executing an action on the machine.
```

Pertempto:
I'll wait a few more minutes. If nothing happens I'll restart my host and try again

Pertempto:
I restarted my host, it seems to work fine now.

docs/vagrant.md
ilearner777: done.
Thanks for recommending the spell checker.  Wanted one, but did not know that it existed.

docs/vagrant.md
ilearner777: I have replaced point 9 with the following:

9. Here is a list of common Vagrant commands. Be sure that you are in the same directory as the Vagrantfile when running these commands!

   |Command|Description|
   |-------|-----------|
   |`vagrant up`|starts vagrant environment (also provisions only on the FIRST vagrant up) Equivalent to pressing the power buttons on your servers.|
   |`vagrant ssh`|connects to machine via SSH|
   |`vagrant suspend`|Suspends a virtual machine (remembers state). Starts faster.|
   |`vagrant halt`|stops the vagrant machine. Similar to shutdown.|
   |`vagrant status`|outputs status of the vagrant machine|

docs/vagrant.md
ilearner777: done.

docs/vagrant.md
ilearner777: done

docs/vagrant.md
ilearner777: Never noticed that button.  Thanks

.vagrant/Vagrantfile
ilearner777: Add this as point 3. in the Caveat section:

1. If you have postgres database running on your host machine at default port `5432`, ensure to shut it down before bring up your VM machine. This can be done by stopping the postgres Windows service. This is because postgres in your VM also listens at the same default port number. Failure to do so will result in the following error:

   ```text
   Vagrant cannot forward the specified ports on this VM, since they
   would collide with some other application that is already listening
   on these ports. The forwarded port to 5432 is already in use
   on the host machine.
   ```

docs/vagrant.md
ilearner777: This is a response to your comment below:

How long does it usually take to boot the VM? This is at the bottom of the output of `vagrant up`.

```text
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
```

I have experienced this issue a number of times as I was working this PR. It appears to be a bug in the relationship between vagrant and virtualbox.  At this point in the process, the VM is up and running and vagrant is trying to establish a SSH session to it. However, it hangs indefinitely at this point and vagrant eventually times out. There is a vagrantfile line `config.vm.boot_timeout` were you can increase the timeout value, however, this would not solve the problem as you would just have to wait long before it times out.  

I don't know why this happens. It happens inconsistently, which makes it hard to debug.

If you find that vagrant is stuck at this point, you can free it up by performing the following steps:
1. Open `Oracle VM VirtualBox Manager`
2. Click your VM machine in the left panel. The name starts with the directory name of the root of your repository. In our case it would be `gloss-translation_....`.
3. If the VM is running, you will see a green right arrow called `Show`. It opens a new panel where you can monitor the boot process of your VM.  For some unknown reason, when I do this, it frees up the vagrant session and then continue booting up to completion.
4. Once the vagrant session is up, you can close down the panel, without shutting down the VM by clicking menu `Machine/Detach GUI`.

This was one of the reasons why I thought we should let developers know that this tool exists.  It does have a lot of useful features in it.

Let me know if you want to add some comment about this issue in the `Caveat` section.

.vagrant/Vagrantfile
Pertempto: Thank you!

docs/vagrant.md
Pertempto: These steps don't work for me because I don't seem to have VirtualBox manager installed.

Pertempto:
My vagrant machine is running fine now, and I can SSH into it. I can not access anything at `localhost:4300`. How do I start the frontend and backend in the machine?

docs/vagrant.md
Pertempto: I'm guessing this will also be true for anyone who has virtualbox installed for them by vagrant. This is probably because we don't need the full suite of virtualbox tools installed to use vagrant.

Pertempto: I think we can leave this be for now. Looking at StackOverflow, it seems like vagrant integration with virtualbox has some common issues. I look forward to using the docker configuration instead. I think it might be the better option since we already use docker in this project for other things.

arrocke:
After you ssh into vagrant, did you start the dev servers?

Pertempto (approved):
Interesting.
Yesterday, I tried using `nx` and `npm` after SSHing in, but neither existed. Today I ran `vagrant up` again (after deleting the `.vagrant` directory when I was in a different branch). Everything worked fine this time.

I'm not sure what was wrong yesterday, but it is working on my end now!

PR #232, opened 2023-10-14
feat: prefill glosses from suggestions

### LINKED ISSUE #209: make suggestions for each word of a verse from approved glosses
## acceptance criteria
* if a word has no gloss for it, suggest the most popular gloss for that lemma form
* suggestions are loaded from new API route - GET /api/language/{code}/verses/{verseId}/suggestions

arrocke:
## What has changed

If there is no gloss, then we are prefilling the input with the top gloss for the lemma form

## Connected Issues

closes #209 

## QA Steps

Assuming that Matthew glosses are empty
1. Translate `en` in Matthew 2:1
2. Navigate to Matthew 2:2
3. Check that `en` is prefilled

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -21,7 +21,7 @@ export interface TranslateWordProps {
   gloss?: string;
   font?: string;
   referenceGloss?: string;
-  previousGlosses: string[];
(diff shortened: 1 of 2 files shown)


Pertempto (approved):
It works great! 👍🏽

PR #234, opened 2023-10-16
feat: show verse translation

### LINKED ISSUE #172: show translation from fetch.bible for the current verse
If we know the preferred translations for the language, we can show the text from the first translation that has the verse. If no translation has the verse, we should notify the user that a translation doesn't exist.

Pertempto:
## What has changed

Added a verse translation line above the glossing area.

## Connected Issues

Closes #172 

## QA Steps

1. Select a few bible translations for a language on the manage language page.
2. Go to the translation page for that language.
    - The translation for the given verse should appear at the top.

## Post-Deployment

- [ ] Nothing required


Pertempto:
The basic concept is ready, but I haven't got to test thoroughly yet. I want to test with a language that has a partial translation, so that I can be sure that the fallback-to-a-different-translation logic works correctly.

Pertempto:
I was quickly able to find a combination in English that demonstrates that the fallback works correctly. Try with a combination of "Text-Critical English New Testament" and "World English Bible". All the OT verse should fallback to World English Bible.

packages/web/src/shared/bibleTranslationClient.ts
```diff
@@ -26,6 +28,41 @@ class BibleTranslationClient {
       name: name_local ? name_local : name_english,
     }));
   }
+
+  async getTranslation(
+    verseId: string,
+    translationIds: string[]
+  ): Promise<{ translationName: string; verseTranslation: string } | null> {
```
arrocke: I prefer to use undefined rather than null in these situations

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -174,6 +178,24 @@ export default function TranslationView() {
     },
   });
 
+  const [translationName, setTranslationName] = useState('');
+  const [verseTranslation, setVerseTranslation] = useState('');
+  useEffect(() => {
+    // Clear the current translation content while loading.
+    setVerseTranslation('');
+    bibleTranslationClient
```
arrocke: We should use `useQuery` instead of `useEffect` since we are dealing with the network. If you don't use something like`useEffect` scenarios like this can happen:

1. Navigate to verse 1
2. Start loading translations for verse 1 - let's say this takes a long time to load due to network issues
3. Navigate to verse 2
4. Start loading translations for verse 2 - this one loads quickly because the network issues are resolved
5. Verse 2 request completes
6. Verse 1 request completes

Since the verse 1 request completes second, it's callback executes second and we end up displaying the wrong verse.

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -290,83 +312,101 @@ export default function TranslationView() {
...
-                isHebrew ? 'ltr:flex-row-reverse' : 'rtl:flex-row-reverse'
-              }`}
-            >
-              {verse.words.map((word, i) => {
-                const targetGloss = targetGlosses[i];
-                const isSaving = glossRequests.some(
-                  ({ wordId }) => wordId === word.id
-                );
+            <>
+              <p
```
arrocke: Once #231 merges, we'll want to set the direction from the language on this element so it matches regardless of the interface language

arrocke (changes requested):

packages/web/src/features/translation/TranslationView.tsx
Pertempto: #231 is ready to merge now 😄

Pertempto:
Why do we have `VerseTranslation` in the prisma schema? Should we remove that, since we're going to dynamically load translations?

Pertempto:
I started trying to use `useQuery`. The loading UX isn't great. There is some bug where the first translation load always errors.

arrocke:
> Why do we have VerseTranslation in the prisma schema? Should we remove that, since we're going to dynamically load translations?

Yep, this can be removed. That was added under the assumption that we would be importing them.

> I started trying to use useQuery. The loading UX isn't great. There is some bug where the first translation load always errors.

Not sure what you need from me here

Pertempto:
> Not sure what you need from me here

It was just a status update, I'll keep working on it today.

packages/web/src/features/translation/TranslationView.tsx
Pertempto: Done. Tested with standard arabic (`arb`)

packages/web/src/shared/bibleTranslationClient.ts
```diff
@@ -61,6 +68,7 @@ class BibleTranslationClient {
         continue;
       }
     }
+    // Must return null, not undefined, so that this will work with useQuery
```
arrocke, line 71: Can you explain why this is the case?

arrocke (commented):
I'd like to reduce the margin a bit. Can we do 16px on top and 8px below?

Also, what do you think about just not showing the paragraph if there is no translation or while it is loading?

packages/web/src/shared/bibleTranslationClient.ts
Pertempto: image

Pertempto:
I fiddled with the spacing a bit, but we NEED to avoid making everything so squished together. It quickly becomes hard to tell at a glance where the translation section ends and the glossing section ends.

It's our natural tendency to give a design too little spacing
- https://www.refactoringui.com/_next/static/media/the-book-1.012ac3f42a15e18e50b5da5328c8c962.png
- https://www.refactoringui.com/_next/static/media/the-book-2.42aba28972115454a8d542bd8399f110.png

packages/web/src/shared/bibleTranslationClient.ts
arrocke: Ah, good to know. That helped me find [this](https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries) in their docs

packages/web/src/shared/bibleTranslationClient.ts
```diff
@@ -26,6 +35,42 @@ class BibleTranslationClient {
...
+        }
+      } catch (e) {
+        console.log(e);
+        // There was some issue getting the verse in this translation, try the
+        // next translation.
+        continue;
+      }
+    }
+    // Must return null, not undefined, so that this will work with useQuery
+    return undefined;
```
arrocke: Do we need to turn this back to `null` for the sake of `useQuery`?

arrocke (commented):
> but we NEED to avoid making everything so squished together

I don't disagree here, but I also don't think our current spacing is guilty of this problem. What you have in this last round of changes looks good. I'm wanting to find a UX designer who can help make these decisions. I have some experience and opinions on this stuff, but not as much as would be useful.

Also, I'm not seeing the code in your last commit that hides the translation line when there isn't a translation. Am I missing something?

packages/web/src/shared/bibleTranslationClient.ts
arrocke: I also need to do some more reading on my own on how typescript thinks about null vs undefined. At my job, we just avoid using null if at all possible and I'm not sure that's right

Pertempto: Interesting. Personally, I avoid `undefined` as much as possible, and find `null` useful, as long as you're careful not to overuse it.

packages/web/src/shared/bibleTranslationClient.ts
Pertempto: Oops. It looks like I somehow missed some changes in the last commit

Pertempto: Fixed

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -68,19 +90,31 @@ function useTranslationQueries(language: string, verseId: string) {
         queryFn: ({ queryKey }) =>
           apiClient.verses.findVerseGlosses(queryKey[2], queryKey[1]),
       });
+      if (selectedLanguage) {
+        queryClient.ensureQueryData({
```
Pertempto:
```suggestion
        queryClient.prefetchQuery({
```

Should I do this instead, like #235?

arrocke: Yep

Pertempto: Done

arrocke (approved):

PR #237, opened 2023-10-20
refactor: use text input for language code

### LINKED ISSUE #222: Rework Language Code Input
The language code input has too many options to show at one time. Instead of using a `ComboboxInput`, let's use a simple input with validation that checks that a given language code is in the official list.

Original background: https://github.com/arrocke/gloss-translation/pull/217#discussion_r1349980119

Pertempto:
## What has changed

Use a text input instead of combo box input for language code.

## Connected Issues

Closes #222 

## QA Steps

1. Try to create a new language, with an invalid language code like "1234".
    - The input should show an error message.
1. Try using a valid language code, like "arb".
    - The language code should be accepted.

## Post-Deployment

- [ ] Nothing required


arrocke (approved):
Nice work!

PR #241, opened 2023-10-26
feat: import BDB definitions

### LINKED ISSUE #163: process to import BDB data
* Use the `BdbMedDef` field from [this data source](https://raw.githubusercontent.com/STEPBible/step/master/step-core-data/src/main/resources/com/tyndalehouse/step/core/data/create/lexicon/lexicon_hebrew.txt) to generate lexicon entries.
* We need a script to process this data and update the database so it can be repeated.
* The bdb definition will need to be converted to markdown

Pertempto:
## What has changed

Added script to insert BDB definitions into the database.

## Connected Issues

Closes #163 

## QA Steps

1. Run `nx run db:import-bdb`
    - All the Hebrew definitions from the BDB should be imported into the `LemmaResource` table
    - Running the command multiple times should not create duplicate entries, because the script first deletes all existing BDB definitions from the table.

## Post-Deployment

- [ ] Run the script for the production DB.

### DIFF
--- a/packages/db/src/scripts/import-bdb.ts
+++ b/packages/db/src/scripts/import-bdb.ts
@@ -42,6 +42,7 @@ const importBdb = async () => {
     });
   await client.$transaction(lemmaUpserts);
   await client.lemmaResource.createMany({ data });
+  console.log('Imported BDB definitions');
 };
 
 console.log('Import BDB definitions...');
--- a/packages/db/src/scripts/import-bdb.ts
+++ b/packages/db/src/scripts/import-bdb.ts
@@ -30,7 +30,7 @@ const importBdb = async () => {
   // resources has something to point to.
   await client.lemma.createMany({ data: lemmaData, skipDuplicates: true });
   await client.lemmaResource.createMany({ data: resourceData });
-  console.log('Imported BDB definitions');
+  console.log('Successfully imported BDB definitions');
 };
 
 console.log('Import BDB definitions...');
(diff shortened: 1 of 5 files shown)


data/lexicon/greek.txt
Pertempto, line 1: I went ahead and included this for future use. Maybe we don't want it.

Pertempto:
Don't forget to add issues for updating our lemma IDs, otherwise many of our new lexicon entries will be inaccessible. https://github.com/arrocke/gloss-translation/issues/163#issuecomment-1779544026

arrocke:
I'll try to take a look at this tomorrow

packages/db/src/scripts/import-bdb.ts
```diff
@@ -0,0 +1,49 @@
...
+        console.error('Missing definition for', lemmaId);
+        return false;
+      }
+      return true;
+    })
+    .map((lemmaId) => {
+      // We have to create non-existent lemmas, so that the foreign key on lemma
+      // resources has something to point to.
+      lemmaUpserts.push(
+        client.lemma.upsert({
```
arrocke: You could do a create many on lemma and set `skipDuplicates: true` to just insert what was missing

arrocke (approved):
Can you generate a new seed file with the BDB data imported?

Pertempto:
> Can you generate a new seed file with the BDB data imported?

Is there any way to do that without completely resetting my database first? I don't want to include all my dummy data.

Pertempto:
I generated a new dump file by:
1. Running the migration reset command
2. Running the seed dump restore command
3. Running `nx run db:import-db`
4. Running the seed dump command

Hopefully the new dump file works correctly.

Pertempto:
The new dump file doesn't reference English by `eng`

Pertempto:
> The new dump file doesn't reference English by `eng`

Fixed

PR #243, opened 2023-10-28
feat: suggestions from google translate

### LINKED ISSUE #210: make suggestions for words from Google Translate
## acceptance criteria
* if a word does not have a gloss, and no approved glosses in other locations use the English gloss to get a translation from Google Translate
* cache Google Translate suggestions so that all words that share the same English gloss can reuse the same suggestion
* add the Google Translate suggestion to the dropdown menu marked with the Google icon
* if the Google Translate suggestion is prefilled mark the text input with the Google icon

Screen Shot 2023-09-27 at 9 41 56 PM

arrocke:
## What has changed

* generates machine translations using Google Translate with the English glosses
* machine translations are cached in the database
* machine suggestions are marked in the user interface with the Google icon

## Connected Issues

closes #210 

## QA Steps

1. Navigate to the translation view
2. Observe glosses marked with the Google icon
3. Approve a gloss
4. Observe approved gloss and Google suggestion in dropdown

## Post-Deployment

- [x] Nothing required


Pertempto:
Can you merge `main` into this branch? I think the db migrations are out of sync, and I don't want to delete my whole DB to run your migrations.

packages/api/shared/machine-translation.tsx
```diff
@@ -0,0 +1,28 @@
+import { TranslationServiceClient } from '@google-cloud/translate';
+
+export class MachineTranslationClient {
+  client: TranslationServiceClient;
+  constructor() {
+    this.client = new TranslationServiceClient({
+      credentials: {
+        client_email: process.env.GOOGLE_TRANSLATE_EMAIL,
+        private_key: process.env.GOOGLE_TRANSLATE_KEY,
```
Pertempto, line 9: I will need these to test?

packages/api/shared/machine-translation.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  async translate(
+    strings: string[],
+    targetLanguage: string
+  ): Promise<string[]> {
+    const [response] = await this.client.translateText({
+      contents: strings,
+      targetLanguageCode: targetLanguage,
+      parent: process.env.GOOGLE_TRANSLATE_PROJECT,
```
Pertempto, line 21: And this

packages/web/tailwind-utilities.js
```diff
@@ -40,18 +40,18 @@ function insetUtilities({ matchUtilities, theme }) {
   matchUtilities(
     {
       start: (value) => ({
-        '[dir="rtl"] &': {
+        '[dir="rtl"] &, &[dir="rtl"]': {
```
Pertempto, line 43: What are these for?

arrocke: Without that the dir attribute would have to be set on an ancestor node for the style to applied. This allows it to be set on the node itself as well.

Pertempto:
We should include code that disables the Google translate logic if the environment variables are missing, otherwise the API throws errors for those missing those variables (which I would assume would include most future devs)

arrocke:
@Pertempto Ok, that's done! This is ready now

Pertempto:
Wow! If I have a Spanish translation line enabled and I see the machine generated glosses, I can almost be a (very low quality) translator for the glosses. I think this will be very powerful for the real translators. 


Notes that don't directly have to do with this iteration:
- Have you considered having more than one translation at the topic of the page? It may be useful to compare different translations at one time. 
- If we don't want to display multiple translation lines at the same time, I think we should at least include some kind of translation switcher, to switch between the different translations enabled for the current glossing language.

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
```diff
@@ -1,13 +1,78 @@
...
+  gloss?: string;
   suggestions: string[];
   state: PrismaTypes.GlossState;
-}[];
+  refGloss?: string;
+  machineGloss?: string;
+}
+
+// This is an optimized sql command for saving the machine glosses for future use.
+// The goal is to save the gloss produced by Google Translate for every word in the database that has the same English gloss.
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts:18]: Why is it based on the English gloss, instead of the original language word?

Pertempto (approved):

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (arrocke) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts:18]: Because Google can't translate from the original languages. We translate from the english glosses, so we can update every word in the database that has the same English gloss regardless of the original language word.

PR #244, opened 2023-10-28
feat: use Noto Sans as interface font

### LINKED ISSUE #239: Standardize fonts
If the gloss fonts are going to be using Noto fonts, we should also use Noto fonts for the interface. We might need to also set fonts for the different interface languages as well

Pertempto:
## What has changed

Use Noto Sans as our default interface font

## Connected Issues

Closes #239

## QA Steps

Use dev tools to inspect various parts of the interface and check that the font in the computed pane is Noto Sans

image

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/tailwind.config.js
+++ b/packages/web/tailwind.config.js
@@ -12,11 +12,10 @@ module.exports = {
     ...createGlobPatternsForDependencies(__dirname),
   ],
   theme: {
-    extend: {
-      fontFamily: {
-        hebrew: ['"SBL-Hebrew"', '"Times New Roman"', 'serif'],
-        greek: ['"SBL-Greek"', '"Times New Roman"', 'serif'],
-      },
+    fontFamily: {
+      sans: '"Noto Sans"',
+      hebrew: ['"SBL-Hebrew"', '"Times New Roman"', 'serif'],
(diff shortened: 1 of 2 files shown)


Pertempto:
This ONLY sets the default interface font. If we want to set different fonts based on currently selected interface language, I'd be open to hearing possible strategies.

packages/web/tailwind.config.js
```diff
@@ -12,11 +12,10 @@ module.exports = {
     ...createGlobPatternsForDependencies(__dirname),
   ],
   theme: {
-    extend: {
-      fontFamily: {
-        hebrew: ['"SBL-Hebrew"', '"Times New Roman"', 'serif'],
-        greek: ['"SBL-Greek"', '"Times New Roman"', 'serif'],
-      },
+    fontFamily: {
+      sans: '"Noto Sans"',
```
arrocke: We should add a fallback fonts for while our font is still loading. Here is how tailwind recommends doing that: https://tailwindcss.com/docs/font-family#customizing-the-default-font

arrocke (changes requested):
Let's tackle picking the right noto font separately

arrocke (approved):

PR #253, opened 2023-10-31
build: improve database seed

arrocke:
## What has changed

I wanted to simplify the database seed so I made a few changes:
* I renamed the seed script to be import-bible. That should only be necessary when we want to change the data we are importing.
* The seed function now just executes the pg_restore command with the database in the environment variable.
* I also configured prisma to execute the seed script after `prisma migrate reset`. Now you can get a fresh database with just one command and you don't have to remember how to use pg_restore.
* I also added `nx run db:seed export` to generate a new seed file as well.

## QA Steps

1. Execute. `nx run db:prisma migrate reset`
2. Log in as Devyn and check that expected data is there. You should see that there are two languages and all of the morphology data

## Post-Deployment

- [x] Nothing required


Pertempto (approved):
Thank you for thinking this through and implementing it! This is a very useful DX improvement!

PR #264, opened 2023-11-09
docs: Add vagrant caveat

### LINKED ISSUE #257: Add Caveat point to vagrant.md
**Describe the bug**
`vagrant up` inconsistently times out during establishment of SSH connection

**To Reproduce**

1. Enter `vagrant up` in terminal session

**Expected behavior**
Expect command to complete without timing out.

- Device:Laptop
- OS: Win11

**Additional context**
Regarding Pertempo comment in PR#225 `chore: add vagrant support`

```text
How long does it usually take to boot the VM? This is at the bottom of the output of vagrant up.

=> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
```

Please confirm if you would like the following comment added to the `Caveat` section of `vagrant.md`.

4. If you experience a long pause at the point in the process shown below. The VM is up and running, but vagrant is trying to establish a SSH session to it. However, sometimes it hangs indefinitely and eventually times out. The cause is unknown.

    ```text
    => default: Waiting for machine to boot. This may take a few minutes...
        default: SSH address: 127.0.0.1:2222
        default: SSH username: vagrant
       default: SSH auth method: private key
    ```

    To free up the process before it times out, perform the following steps:

        1. Open `Oracle VM VirtualBox Manager`.
        2. Click on the name of your VM machine in the left panel. The name starts with the name of the root folder of your repository. In our case it would be `gloss-translation_....`
        3. If the VM is running, you will see a green right arrow called `Show`. It opens a new panel where you can monitor the boot process of your VM. This should free up the vagrant session and then the booting up should continue to completion.
        4. Once the vagrant session is up, you can close down the panel, without shutting down the VM, by clicking menu` Machine/Detach GUI`.

If agree with this suggestion, please assign to me for implementation.

ilearner777:
## What has changed
Added a new point to the `Caveat` section of document `vagrant.md`. 
This addition explains how to recover from a `vagrant up` command that gets hung during establishment of a SSH session.

## Connected Issues

closes #257 

## QA Steps
Review document changes and ensure that they faithfully reflect what has been agreed upon in issue #257.

NOTE: I have made a few minor changes for clarity. 

## Post-Deployment

- [X ] Nothing required


arrocke (changes requested):
Looks like there are some extra files in here that need to be removed

Pertempto:
@ilearner777 You can add these files to your user-specific gitignore, so that you don't have to keep manually removing them before committing them: https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer


In the case of these files, it will take a little extra work to make git ignore them, since you already added them in the past:
https://stackoverflow.com/a/1139797

ilearner777:
> @ilearner777 You can add these files to your user-specific gitignore, so that you don't have to keep manually removing them before committing them: https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
> 
> In the case of these files, it will take a little extra work to make git ignore them, since you already added them in the past: https://stackoverflow.com/a/1139797

Thankyou for taking the time to offer me these tips. :)

ilearner777:
Extra files removed

arrocke (approved):

PR #269, opened 2023-11-11
fix: next button arrow

arrocke:
## What has changed

In hebrew, the arrow in the next button was in the wrong place. I think this should fix that.

## QA Steps

Look at the next button at the end of the translation line for Genesis and Matthew in English and Arabic interface languages. The arrow should point right for greek and left of hebrew regardless of interface language.

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslationView.tsx
+++ b/packages/web/src/features/translation/TranslationView.tsx
@@ -416,13 +416,13 @@ export default function TranslationView() {
                         );
                       }}
                     >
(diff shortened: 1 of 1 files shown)


Pertempto:
I need to load the manage language page to assign my dummy account as a translator, but it isn't loading. Messaged you about that on Slack. Can't test until that is resolved.

Pertempto (approved):

PR #270, opened 2023-11-11
refactor: constrain max width of site

arrocke:
## What has changed

One large screens, everything is going to get very spread out. Most sites set a max width for the main content on the site. I've set this to 1500px so there is plenty of room for the translation line and sidebar.

## QA Steps

Use the responsive device view in the browser dev tools to test on a really wide screen

## Post-Deployment

- [x] Nothing required


Pertempto:
Your description says 1500px but the code says 1800px. Which one do we want?

Pertempto (approved):
It looks good on my 2k monitor at 1800px width

PR #278, opened 2023-11-17
docs: vagrant plugin for file events

### LINKED ISSUE #273: Webpack Dev Server Auto Reload Does Not Work in Vagrant
**Describe the bug**
The auto-reload feature of the webpack dev server doesn't seem to work when running inside Vagrant.

**To Reproduce**

1. Run `nx serve web` from inside your vagrant machine (it can use the Virtualbox or Docker provider, both have the bug)
2. Open a web browser and go to `localhost:4200`
    - Verify that the web page loads correctly
3. Manually replace `{t('common:app_name')}` with `This is a test` in `Header.tsx`

**Expected behavior**
The change should automatically propagate to the web browser. The title in the header should show "This is a test" instead of "Gloss Translation"

- Device: Laptop
- OS: Windows 11
- Browser: Vivaldi (Chromium-based)
- Vagrant: 2.3.7

**Additional context**
This bug occurs for both the Virtualbox and Docker providers.

arrocke:
## What has changed

Adds vagrant step to configure plugin to forward file change events.

## Connected Issues

closes #273 

## QA Steps

1. Run `vagrant plugin install vagrant-notify-forwarder`
2. Run `vagrant reload`
3. Start vagrant and the dev servers
4. Change a file and confirm it reloads

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -58,7 +58,7 @@ Vagrant.configure("2") do |config|
   config.vm.network "forwarded_port", guest: 5432, host: 5432, host_ip: "127.0.0.1"
 
   # Mount repo project root folder. 
-  config.vm.synced_folder ".", "/home/vagrant/vmrepo"
+  config.vm.synced_folder ".", "/home/vagrant/vmrepo", fsnotify: true
 
   # Disable the default share of the current code directory. Doing this
   # provides improved isolation between the vagrant box and your host
--- a/docs/vagrant.md
+++ b/docs/vagrant.md
@@ -40,6 +40,7 @@ This was tested using:
    - [VirtualBox](https://www.virtualbox.org/)
    - [Docker Desktop](https://docs.docker.com/desktop/)
 
+1. Install the notify forwarder plugin with the command `vagrant plugin install vagrant-notify-forwarder`
 1. If you have not already done so, fork repo to your local machine.
 1. Vagrant uses a special configuration file called `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host). It is located in your repo root directory.
 1. To start the creation of your virtual environment, run one of these commands in the repo root directory. (This takes a while so be patient.)
--- a/docs/vagrant.md
+++ b/docs/vagrant.md
@@ -40,7 +40,7 @@ This was tested using:
    - [VirtualBox](https://www.virtualbox.org/)
    - [Docker Desktop](https://docs.docker.com/desktop/)
 
-1. Install the notify forwarder plugin with the command `vagrant plugin install vagrant-notify-forwarder`
+1. Install the notify forwarder plugin with the command `vagrant plugin install vagrant-fsnotify`
 1. If you have not already done so, fork repo to your local machine.
 1. Vagrant uses a special configuration file called `Vagrantfile` that contains all the information required for the creation of a Virtual Machine (guest) on your workstation (host). It is located in your repo root directory.
 1. To start the creation of your virtual environment, run one of these commands in the repo root directory. (This takes a while so be patient.)
@@ -55,6 +55,8 @@ This was tested using:
    ```bash
    vagrant ssh
    ```
+
+1. In order for the dev servers to be able to react to file changes, run `vagrant fsnotify`.
 
 1. Here is a list of common Vagrant commands. Be sure that you are in the same directory as the Vagrantfile when running these commands!
 


ilearner777:
## Environment
- Laptop
- Windows 11 home edition

## Steps performed
1. vagrant plugin install vagrant-notify-forwarder
2. vagrant up

## Error Encountered

```text
    default: Notify-forwarder: guest listening for file change notifications on 0.0.0.0:29324.
==> default: Notify-forwarder: Unsupported host operating system
```

## Cause
Only the following Operating Systems are supported
- FreeBSD 64 bit as guest,
- Linux 64 bit as host and guest, and
- Mac OS X 64 bit as host and guest.

f you're running an unsupported host or guest and want to disable this plugin for a specific machine, add the following line to your Vagrantfile:

```ruby
config.notify_forwarder.enable = false
```
## Recommendation
Add a note about this restriction in the documentation

Pertempto:
We need something that works on Windows, not just a comment about it in the documentation. We don't want to cut out developers running on Windows, otherwise why even have vagrant?

ilearner777:
so far I have tried everything that I can find online and have not been able to get it to work in windows. will post later what I have tried to see if anyone can see if I have done anything wrong.

ilearner777:
I have done some research that may explain why I have not been able to get hmr to work in our VM.

## Issue
Vagrant does not implement inotify that is required for hmr to work.

## Recommend Solution
Implement watch option polling

## Information Discovered
1. Webpack 5 supports watch option polling. reference [webpack watch options](https://webpack.js.org/configuration/watch/). There you will see the following sample code:

```javascript
module.exports = {
  //...
  watchOptions: {
    aggregateTimeout: 200,
    poll: 1000,
  },
};
```
2. We use  `@nx/webpack: 16.7.1,`.  Here is a link to the support options: [@nx/webpack options](https://nx.dev/nx-api/webpack/executors/dev-server).   There they list all available options.  I could not find any watch options that support polling. These are the supported options:
```json
// "executor": "@nx/webpack:dev-server", 
// "options": 
{
  "buildTarget": "<string>",
  "port": 18,
  "host": "<string>",
  "ssl": true,
  "sslKey": "<string>",
  "sslCert": "<string>",
  "watch": true,
  "liveReload": true,
  "hmr": true,
  "publicHost": "<string>",
  "open": true,
  "allowedHosts": "<string>",
  "memoryLimit": 55,
  "baseHref": "<string>"
}
```

## Copilot Conclusion

Regarding the polling option, it's not directly available in the @nx/webpack:dev-server executor options. However, it can be set in the underlying webpack configuration. If you're facing issues with file watching in a virtualized environment like Vagrant, you might need to manually adjust your webpack configuration to enable polling.

## Next Steps

Looks like we need to see if there is a way to implement polling for file watch in our project. Copilot has recommended the following. Let me know if this is something that could be implemented in our project. If so, modify the code below for our case and I will give it a try.

## Copilot Recommendation

To enable polling for file watching in your webpack configuration when using the `@nx/webpack:dev-server executor`, you would need to extend the webpack configuration used by Nx.

First, create a `webpack.config.js` file in the root of your workspace (if it doesn't already exist). In this file, you can export a function that takes the default configuration as an argument and returns the modified configuration.

Here's how you can enable polling in your `webpack.config.js`:

```text
module.exports = (config, context) => {
  // Check if we're in dev-server mode
  if (context.target.target === 'serve') {
    config.watchOptions = {
      poll: 1000, // Check for changes every second
      aggregateTimeout: 600, // Delay the rebuilt after the first change
    };
  }

  return config;
};
```

This configuration will check for changes every second and delay the rebuilt after the first change.

Next, you need to tell Nx to use this custom webpack configuration. In your `workspace.json` (or angular.json), find the project and target that uses the @nx/webpack:dev-server executor. Add a `customWebpackConfig` option that points to your `webpack.config.js` file:

```json
"architect": {
  "serve": {
    "builder": "@nrwl/web:dev-server",
    "options": {
      "buildTarget": "your-project:build",
      "customWebpackConfig": {
        "path": "./webpack.config.js"
      }
    }
  }
}
```
Replace "your-project:build" with the actual build target for your project.

Now, when you run nx serve your-project, Nx will use your custom webpack configuration, which enables polling for file watching.

Pertempto:
I agree that polling might be the way to go.

arrocke:
Sorry, I didn't realize that plugin didn't support Windows.

I've spent a few hours between yesterday and today trying to get polling to work, but I can't figure it out. [This](https://webpack.js.org/configuration/dev-server/#devserverwatchfiles) is the config that should turn that on, but it doesn't work when I try it. We can't use `watchOptions` as suggested above because that is for webpack watch, but nx uses the webpack dev server.

I did find a different plugin that should work in all systems, if you want to test that:
1. Update branch
2. Run `vagrant plugin uninstall vagrant-notify-forwarder`
3. Run `vagrant plugin install vagrant-fsnotify`
4. Run `vagrant reload`
5. Run `vagrant fsnotify`
6. Start dev servers

ilearner777:
I tried the new plugin. 
Found out that you need to modify a line to the Vagrantfile to get it to work:

```ruby
config.vm.synced_folder ".", "/home/vagrant/vmrepo",fsnotify: true 
```

After that the plugin started working.  
However, even with that, the hmr didn't update the site; not even with a page refresh.
The `fsnotify` was working as you can see in the log output below


Screenshot_1

Not sure if Addison wants to give it a try.

If it doesn't work for him, we may be out of options to get this working.
I can't think of anything else.

Let me know if you want to just put it on hold or back out the vagrant support feature from our repo.

Pertempto:
I can try Monday, busy this evening

Pertempto:
I sadly got the same results as @ilearner777. I can see the fsnotify events triggered when I update files, but webpack doesn't acknowledge the updates at all, and nothing changes in the browser.

@arrocke how difficult would it be to convert to using vite instead of webpack? I'm guessing it would be non-trivial, but webpack seems to be the root of our issues here. Vite is quite a bit more performant than webpack. Angular recently switched from using webpack to vite for their build system, and we've seen great performance gains in build time at work.

Pertempto:
I take that back. After doing more research it seems the problem has more to do with virtual machines having poor support for file system events. It seems like this should be a solved problem. Don't many development teams use virtual machines for consistency? Surely they use hot-reload too.

Pertempto:
Should we try using docker directly for our VMs, instead of using vagrant?

https://stackoverflow.com/questions/42445288/enabling-webpack-hot-reload-in-a-docker-application

Pertempto:
I experimented with using `inotify-tools` to debug the filesystem events. I am pushing a commit to this branch that adds the install to the `vagrantfile`

1. `git pull`
1. `vagrant provision --provision-with inotify-tools`
3. `vagrant fsnotify`

In a separate terminal:
1. `vagrant ssh`
2. `inotifywait -m packages/web/src/app/Header.tsx`

In a separate terminal:
1. `vagrant ssh`
2. `nx serve web`


After the watchers are setup, try modifying `packages/web/src/app/Header.tsx`

These are the results I get:
1. `vagrant fsnotify` reports the change
2. `inotifywait` from inside the VM reports the changes
4. Webpack does NOT report the changes, it does NOT update anything


So maybe the problems are rooted in webpack?



## Edit

If you try refreshing the browser, then webpack DOES detect the changes. So it seems now that the only problem is that our browser doesn't automatically get changes from webpack. It would be great to figure that out, but I don't think it is a deal breaker to have to manually refresh the browser to get the newest version of the app.

Maybe webpack isn't correctly hooked up with the filesystem events, and just checks the filesystem when it receives a new request triggered by refreshing the browser?

ilearner777:
Great to hear you are making progress.

Re web browser updating, it appears that webpack depends on websocket for this function.  Not sure if there is a disconnect between the browser and webpack because it has to go through VM gateway.  Not sure any special configuration adjustment needs to be made.

ilearner777:
On another point. when I reboot my laptop, I have to manually restart postgresql. Has anyone else experienced this problem?

Pertempto:
What do you mean by "restart postgresql"?

It should be just running in your virtual machine so it's available after you run `vagrant up`

Pertempto:
> Great to hear you are making progress.
> 


Did u try my steps? Did they work for you? If so, I think I'll remove inotify tools because we don't need it long term, then we can merge this PR.

arrocke:
> On another point. when I reboot my laptop, I have to manually restart postgresql. Has anyone else experienced this problem?

This is unrelated, that's a good question for the #dev channel in slack

> I sadly got the same results as @ilearner777. I can see the fsnotify events triggered when I update files, but webpack doesn't acknowledge the updates at all, and nothing changes in the browser.

> If you try refreshing the browser, then webpack DOES detect the changes. So it seems now that the only problem is that our browser doesn't automatically get changes from webpack.

What I observed is that it took upwards of 15 seconds for webpack to recompile and update in the terminal. Everything works as expected, it just takes a lot longer.

At this point, I think we have two options:
1. We merge this PR with the `fsnotify` plugin and move on.
2. We abandon vagrant completely and remove it from the repo

For me, the performance and storage requirements of a VM make vagrant unusable on my 2015 macbook which I do most of the development on this project. In this case, performant development outweighs a standardized development environment. In my opinion, the set up of this project is not terribly complicated, (setting aside the import lambda, which will be removed in a few months), so the benefits from vagrant are minimal. I'm personally planning to continue without vagrant, and I don't want to spend much more time trying to get support for it. I think it was worth trying, but there are more important things to work on.

Pertempto:
That is a fair assessment

Pertempto:
I'll get posgres running again on my host machine and quit using vagrant

arrocke:
> I'll get posgres running again on my host machine and quit using vagrant

Does that mean you are voting for option 2?

Pertempto:
Yes

Pertempto:
If you're not going to use vagrant then the rest of us shouldn't. It's not worth it

arrocke:
Ok, going to close this and put an issue in to remove vagrant. Thank you @ilearner777 for all you work on this

Issue #248, opened 2023-10-30
Dealing with different lemma id systems


arrocke:
**Is your feature request related to a problem? Please describe.**
The data we have for strongs numbers for each word doesn't have the extra strongs numbers that were added later.

**Describe the solution you'd like**
We should see if the data in the step bible repo can get us the updated strongs numbers

**Describe alternatives you've considered**
If necessary we can do this manually. I'm not sure how many of these there are.

arrocke:
We can use these two data sources to get the disambiguated strongs numbers:
https://github.com/STEPBible/STEPBible-Data/tree/master/Translators%20Amalgamated%20OT%2BNT

Reference: https://tinyurl.com/DisambiguatedStrongs

arrocke:
Ok, I think I have some more clarity on this one. OpenScriptures introduced new strongs numbers and extended several numbers with letters to account for the additional entries in BDB that weren't in strongs concordance. An example:

Strongs H3588 is just כִּי which is a connecting word. However, BDB has two additional entries כִּי אם־ and כִּי עַל כֵּן that strongs doesn't have. So OpenScriptures marks what strongs has as H3588a and the two additional BDB entries as H3588b and H3588c.

I think at this point we should convert our lemma IDs to use the extended strongs from OpenScriptures and then update the lemma IDs for each word in the database to match. OpenScriptures has [XML files](https://github.com/openscriptures/morphhb/tree/master/wlc) with that data.

arrocke:
Andrew suggested that we use unfolding word's dataset [here](https://git.door43.org/unfoldingWord/hbo_uhb). Their's is based off of OpenScriptures, but they've invested time and money to make it more reliable.

PR #286, opened 2023-11-21
feat: updated LSJ entries

### LINKED ISSUE #277: Migrate to new LSJ datasource
Our current datasource for the LSJ is protected by a BSD license

We should rebuild the import to use [this datasource](https://raw.githubusercontent.com/STEPBible/STEPBible-Data/master/Lexicons/TFLSJ%20extra%20-%20Translators%20Formatted%20full%20LSJ%20Bible%20lexicon%20-%20STEPBible.org%20CC%20BY.txt) instead.

**Additional context**
We will also need to sort out which words come from Bill Mounce's lexicon and either filter them out, or get permission to use.
We should also credit step bible in our project readme and potentially in the app as well

Pertempto:
## What has changed

- Added data source section to `README`
- Updated greek data file
- Updated LSJ import logic to support the new data file
- Modified BDB import logic
- Updated seed file

## Connected Issues

Closes #277 

## QA Steps

- `nx run db:import-lsj`
- In database: `SELECT * FROM "LemmaResource" WHERE "resourceCode" = 'LSJ';`
    - NOTE: the definitions are multi-line. Click on one to see the whole entry.

## Post-Deployment

- [ ] Run `nx run db:import-db` for production DB

### DIFF
--- a/packages/db/src/scripts/helpers/parse-lexicon.ts
+++ b/packages/db/src/scripts/helpers/parse-lexicon.ts
@@ -1,36 +1,44 @@
 import { createReadStream } from 'fs';
 import { createInterface } from 'readline';
 
-/**
- * This is used to parse our lexicon files, so that they can be inserted into
- * the DB.
- */
-export const parseLexicon = async (
-  filename: string,
-  keys: string[]
-): Promise<Record<string, Record<string, string>>> => {
-  const input = createReadStream(filename);
-  const reader = createInterface({ input });
-  const parsed: Record<string, Record<string, string>> = {};
+export const parseBdb = async (filename: string) => {
+  const parsed: Record<string, string> = {};
   let currentId = '';
-  let currentData: Record<string, string> = {};
-  for await (const line of reader) {
+  await lineByLine(filename, (line) => {
     const indicator = line[0];
     const rest = line.substring(1);
-    if (indicator == '$') {
-      if (currentId) {
-        // Record the last data before overwriting the temporary variables.
-        parsed[currentId] = currentData;
-      }
-      currentData = {};
-    } else if (indicator == '@') {
+    if (indicator == '@') {
       const [key, value] = rest.split('=\t', 2);
-      if (keys.includes(key)) {
-        currentData[key] = toMarkDown(value);
+      if (key == 'BdbMedDef') {
+        parsed[currentId] = toMarkDown(value);
       } else if (key == 'StrNo') {
         currentId = value;
       }
     }
+  });
+  return parsed;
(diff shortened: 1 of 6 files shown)


Pertempto:
@arrocke What did you mean by 
> We will also need to sort out which words come from Bill Mounce's lexicon and either filter them out, or get permission to use.

I didn't do anything with that

data/seed.dump
Pertempto, line 1: It's strange that this file went down so much in size. I reset my DB to the seed file, and the entries are included.

arrocke:
> We will also need to sort out which words come from Bill Mounce's lexicon and either filter them out, or get permission to use.

I couldn't find a way to identify which entries are from his lexicon. Let's ignore for now

arrocke (approved):

packages/db/src/scripts/helpers/parse-lexicon.ts
```diff
@@ -1,36 +1,44 @@
...
     }
+  });
+  return parsed;
+};
+
+export const parseLsj = async (filename: string) => {
+  const parsed: Record<string, string> = {};
+  await lineByLine(filename, (line) => {
+    const split = line.split('\t');
+    const id = split[1].split('=')[0].trim();
```
arrocke: I think we are going to be using the eStrongs number across the board. In Hebrew that aligns with the BDB. #248 should give us a datasource that aligns with eStrongs so we'll be able to update each word to match these LSJ entries. Can you modify this to use that as the strongs?

arrocke (changes requested):
Before this get's merged, let's merge #288 and then rebuild this seed with the LSJ data.

packages/db/src/scripts/helpers/parse-lexicon.ts
Pertempto: Then how should I handle multiple entries for the same eStrongs number?

image

Pertempto: I was using the `dStrong` column so that we could differentiate between these entries.

arrocke: For now, let's take the first entry. dStrongs come from StepBible and are used to separate spellings and submeanings of words, and it's not clear to me whether that makes sense for us. I'm still working with Andrew to wrap my head around all of this

Pertempto: Done. Also improved parsing of bold/italics indicators.

arrocke (approved):

PR #288, opened 2023-11-21
fix: import hebrew words without strongs numbers

### LINKED ISSUE #285: missing words in Hebrew
**Describe the bug**
There are words in the current app that aren't in this one.

**To Reproduce**
Comare Obadiah 5 in hebrewgreekbible.online and globalbibletools.com. There are words missing.

**Expected behavior**
These missing words should be there.

**Additional context**
This appears to be caused by how we imported data from the old app. Our filter for which elements in the data file filtered out Hebrew words with no lemma ID

arrocke:
## What has changed

Previously, we were only using words in Sheldon's dataset that had 6 entries, but this excludes some words in the Hebrew text that don't have strongs numbers. This PR removes that filter, and updates the data seed.

I didn't bother to readd any of the lexicons back to the seed since those are still in flux.

## Connected Issues

closes #285 

## QA Steps

Import the seed and check Obadiah 5 for correctness.

## Post-Deployment

I've already updated the data in production

### DIFF
--- a/packages/db/src/scripts/import-bible.ts
+++ b/packages/db/src/scripts/import-bible.ts
@@ -55,12 +55,9 @@ async function run() {
           },
         });
 
-        // Some entries in the words list are text in between words, so we ignore them.
-        const filteredWords = words.filter((word) => word.length === 6);
-
-        for (let wordIndex = 0; wordIndex < filteredWords.length; wordIndex++) {
-          const [text, , english, grammar, rawStrongs] =
-            filteredWords[wordIndex];
+        for (let wordIndex = 0; wordIndex < words.length; wordIndex++) {
+          const [text, , english, grammar = '', rawStrongs = '????'] =
+            words[wordIndex];
 
           // We clean the strongs codes since the hebrew ones have some extra characters.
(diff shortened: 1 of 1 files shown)


packages/db/src/scripts/import-bible.ts
```diff
@@ -55,12 +55,9 @@ async function run() {
...
         });
 
-        // Some entries in the words list are text in between words, so we ignore them.
-        const filteredWords = words.filter((word) => word.length === 6);
-
-        for (let wordIndex = 0; wordIndex < filteredWords.length; wordIndex++) {
-          const [text, , english, grammar, rawStrongs] =
-            filteredWords[wordIndex];
+        for (let wordIndex = 0; wordIndex < words.length; wordIndex++) {
+          const [text, , english, grammar = '', rawStrongs = '????'] =
```
arrocke, line 59: For now, words without a strongs number are linked to a single lemma, but in #248, I'm working on getting access to a datasource that can distinguish these words and their grammatical forms since Sheldon's data didn't have this.

Pertempto (approved):

PR #298, opened 2023-12-04
feat: reorder translations

### LINKED ISSUE #208: Add and reorder translation preferences
## acceptance criteria
* user needs to be able to add/remove reference translations
* user needs to be able to reorder translations according to their preference

Pertempto:
## What has changed

- Added `ReorderableMultiselectInput`
- Used `ReorderableMultiselectInput` for translation selector on the manage language page

## Connected Issues

Closes #208

## QA Steps

1. Go to the manage language page for English
2. Add "World English Bible"
3. Add "Wycliffe's Bible with Modern Spelling (Enhanced)"
4. Save changes
5. Go to Malachi 4:6
    - The translation should use World English Bible
6. Go to Matthew 1:1
    - The translation should use World English Bible
1. Go to the manage language page for English
2. Move "Wycliffe's Bible with Modern Spelling (Enhanced)" above "World English Bible"
3. Save changes
4. Go to Malachi 4:6
    - The translation should use Wycliffe's Bible
5. Go to Matthew 1:1
    - The translation should use World English Bible

## Post-Deployment

- [ ] Nothing required


packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+                    onClick={() => removeItem(i)}
+                  >
+                    <Icon icon="close" />
+                    <span className="sr-only">{t('common:close')}</span>
+                  </button>
+                </div>
+              );
+            })}
+        </div>
+        <Form
```
Pertempto: @arrocke, I need help with this. How should I implement the add translation selector?

At first I tried to use a controlled `ComboboxInput`. It didn't seem to work, and I think that it was because the `ComboboxInput` was picking up the outer form context of the manage language form.

Then I tried to use a `Form` here, but I get this:
```
Warning: validateDOMNesting(...): <form> cannot appear as a descendant of <form>.
```


How can I implement the new translation selector inside my input component, without clashing with the outer form?

arrocke: Ah, maybe it wasn't the best idea to have our base components connect to Form components by default if they are nested inside. That makes them less flexible. Maybe we add a boolean prop to the ComboboxInput that defaults to false named `disconnect` or `isolate` that if true, prevents connection to the formContext

Pertempto: Ok, that worked. Now I'm getting this annoying warning

```
Warning: Received `true` for a non-boolean attribute `isolate`.
```

We don't get this warning for `fitToScreen` in `View`. What about my `isolate` prop is causing this warning? It IS set to a boolean type (though I'm guess that doesn't matter because TypeScript is only ran at compile time).

Pertempto:
This should be ready!

packages/web/src/shared/components/Icon.tsx
```diff
@@ -21,5 +21,8 @@ library.add(
   FaSolid.faArrowRight,
   FaSolid.faRightFromBracket,
   FaSolid.faExclamationTriangle,
-  FaSolid.faFileImport
+  FaSolid.faFileImport,
+  FaSolid.faGripVertical,
```
arrocke: I think we can remove the grip icon for now

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
```diff
@@ -0,0 +1,152 @@
...
+import Button from '../actions/Button';
+import ComboboxInput from './ComboboxInput';
+
+export type ReorderableMultiselectInputProps =
+  BaseReorderableMultiselectInputProps & {
+    required?: boolean;
+    isolate?: boolean;
+  };
+
+export default function ReorderableMultiselectInput(
```
arrocke: Can we use Sortable instead of Reorderable in the name?

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
```diff
@@ -0,0 +1,152 @@
...
+                    <Icon icon="close" />
+                    <span className="sr-only">{t('common:close')}</span>
+                  </button>
+                </div>
+              );
+            })}
+        </div>
+      )}
+      <div className="flex gap-1">
+        <ComboboxInput
```
arrocke: We should attach the ref from `forwardRef` to the combobox input, so that react-hook-form can focus the list box element.

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
```diff
@@ -0,0 +1,152 @@
...
+  onChange?(value: string[]): void;
+  onBlur?(): void;
+}
+
+type ItemType = { label: string; value: string };
+
+const BaseReorderableMultiselectInput = forwardRef<
+  HTMLInputElement,
+  BaseReorderableMultiselectInputProps
+>(({ className = '', value, onChange, items }, ref) => {
```
arrocke: We need to add onBlur here so that react-hook-form can tell when this input has been touched. Maybe we can call onBlur when any child element inside the component blurs

arrocke (changes requested):
This looks really great! A couple things to address yet

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
Pertempto: Do I need to do some modifications to `ComboboxInput` to do this? I'm getting an error about it not having a `ref` prop.

packages/web/src/shared/components/Icon.tsx
Pertempto: Done

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
Pertempto: Done

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
Pertempto: Done. Will this cause issues when the user tabs between the different children elements? `onBlur` will be called multiple times, while the entire component still has focus.

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
Pertempto: @arrocke did you see this?

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
arrocke: Ah yeah. We can leave this off for now since we don't have validation on this component

packages/web/src/shared/components/form/ReorderableMultiselectInput.tsx
arrocke: Found the issue. In the ComboboxInput component, `props.isolate` is passed to the base component. When that happens, it is also passed to `Combobox.Input`. The error in the browser is because react is trying to set isolate on the input element as an HTML attribute. So if we destructure the props in the component so we don't pass it down to child components, we should be good.

arrocke (approved):
Once that warning is taken care of, we can merge this

PR #300, opened 2023-12-20
chore: import unabridged BDB

### LINKED ISSUE #280: Switch to unabridged BDB
We can't use the abridged BDB since that is under copyright. The full BDB is what we want to translate anyway.

Use the full BDB for each lemma and remove the abridged. We'll need to do some work to align the BDB to our lemmas, or modify our lemmas to align with the BDB

**Additional context**
[This source](https://github.com/eliranwong/unabridged-BDB-Hebrew-lexicon/tree/master) has the full BDB in CSV and JSON formats and an alignment to strongs numbers
[This source](https://github.com/openscriptures/HebrewLexicon/tree/master) has another alignment between BDB and strongs, but doesn't have a complete BDB

arrocke:
## What has changed

* Render lexicon entries as sanitized HTML rather than Markdown. The BDB has super and subscripts which don't have Markdown syntax. And we need to be able to mark text as RTL within a sentence.
* Update the LSJ import to convert to HTML
* Update the BDB import to use the unabridged BDB aligned to strongs numbers
* Wrap greek and hebrew in lexicon entires in spans with classes to be able to set the appropriate font
* Remove the strongs concordance import
* Expanded the sidebar to make it more readable on wider screens

## Connected Issues

closes #280 

## QA Steps

1. Run `DELETE FROM "LemmaResource"` in the database
2. Run `nx run db:import-bdb` and `nx run db:import-lsj`
3. Load a verse from the OT and NT
4. Check lexicon entries for formatting and fonts

## Post-Deployment

I'll replace the LemmaResources table with the correct data


Pertempto:
Are we missing a dependency?
```
TSError: ⨯ Unable to compile TypeScript:
packages/db/src/scripts/import-bdb.ts(4,27): error TS2307: Cannot find module 'fast-xml-parser' or its corresponding type declarations.
packages/db/src/scripts/import-bdb.ts(5,10): error TS2305: Module '"@prisma/client"' has no exported member 'BDBEntry'.
```

arrocke:
One of those is because a new dependency was added to parse xml. The other was a remnant from my messy testing branch where I was importing stuff into the database to figure out the data. Should be good now if you run `npm i`

Pertempto:
The other problem was that `fast-xml-parser` wasn't specified in our `package.json`. I added it to dev dependencies.

Pertempto:
Some BDB entries are almost entirely wrapped in `<strong>` (See the first words of Genesis 1:2 or Malachi 4:6)

Pertempto:
On the LSJ side, we have these helpful debug messages:
```
Importing LSJ definitions...
Parsed 5524 words
Successfully imported LSJ definitions
```

Could we add them to the BDB import?

Pertempto:
The sidebar width seems to be changing when I switch verses. It should stay static.

Try switching from Romans 15:14 (narrow sidebar) to Romans 15:15 (wide sidebar) to Romans 15:15 (narrow sidebar).


It might have to do with the length of content from the Bible translation. For reference, I'm testing with _Palabra de Dios para ti_ in Spanish.


image
image

Pertempto (changes requested):
See above comments.

arrocke:
Great stuff. I've made the following changes. You'll want to delete your BDB data and reimport before you test.
* Prevent the sidebar from shrinking when the translation view wants to expland
* Fixed the import to handle a few edge cases with the source data that caused unnecessary bold text
* Added some logs to the bdb import script
* Standardized the line height in the lexicon entry to account for the different languages

Pertempto (approved):
It works well! I'll be unavailable until Tuesday, here a few remaining questions I had:
- Are we going to update the seed dump file?
- The old import BDB process deleted all BDB entries before importing the new entries. Why not retain that behavior? What is the benefit of having to delete manually?
    - It is confusing when the import-bdb process hangs indefinitely when you forget to delete the entries beforehand.

arrocke:
I addressed both of those questions

PR #312, opened 2024-01-03
feat: make gloss approval more discoverable

### LINKED ISSUE #282: make gloss approval more discoverable.
In all user interviews I've done, users aren't discovering that glosses can be approved and unapproved. Additionally, there is no way to approve/unapprove glosses on a mobile device (including tablets) since we can't listen for Enter on them.

One option is to add a button that's not in the tab sequence after each input to either approve or unapprove the gloss. We can additionally only make it appear when the input is focused, and also put the keyboard shortcut in a tooltip.

Pertempto:
## What has changed

Added approve and revoke approval buttons.

## Connected Issues

Closes #282 

### Approve Button
1. Go to the translate page.
2. Focus an unapproved gloss input that contains content
    - A green approval button should appear
    - It should not be in the tab sequence
    - It should have a localized tooltip
    - Clicking it should approve the gloss

### Revoke Button
1. Go to the translate page.
2. Focus an approved gloss input
    - A red revoke button should appear
    - It should not be in the tab sequence
    - It should have a localized tooltip
    - Clicking it should remove approval from the gloss

Also test with machine generated glosses, to make sure that the icon is positioned correctly.
## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -66,6 +66,7 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
     );
 
     const glossValue = gloss || suggestions[0] || machineGloss;
+    const [currentInput, setCurrentInput] = useState(glossValue ?? '');
     const hasMachineSuggestion = !gloss && !suggestions[0] && !!machineGloss;
 
     const glossWidth = useTextWidth({
@@ -137,7 +138,7 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
               dir={targetLanguage?.textDirection ?? TextDirection.LTR}
             >
               <div className="group-focus-within/input-row:block hidden">
-                {glossValue && status !== 'approved' && (
+                {currentInput && status !== 'approved' && (
                   <Button
                     className="!bg-green-600"
                     tabIndex={-1}
@@ -221,6 +222,9 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
                         approved: !implicit && !!value,
                       });
                     }
+                  }}
+                  onInput={(event) => {
+                    setCurrentInput((event.target as HTMLInputElement).value);
                   }}
                   onKeyDown={(e) => {
                     if (e.metaKey || e.altKey || e.ctrlKey) return;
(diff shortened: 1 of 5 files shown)


Pertempto:
# To Do

- [x] Fix styling of machine gloss icon

packages/web/src/public/locales/en/translate.json
```diff
@@ -1,9 +1,11 @@
...
+  "approve_tooltip": "Approve (Enter)",
   "approved": "Approved",
   "interlinear": "Interlinear",
   "next_verse": "Next Verse",
   "previous_verse": "Previous Verse",
   "saving": "saving",
   "select_verse": "Select verse",
-  "strongs": "Strongs"
+  "strongs": "Strongs",
+  "unapprove_tooltip": "Unapprove (Esc)"
```
Pertempto: Should we use a different term? "Unapprove" isn't a real word, so it doesn't translate well.

Pertempto:
Ok, this should be ready. I think the machine gloss icon is positioned correctly now.

packages/web/src/public/locales/en/translate.json
arrocke: Let's use "Revoke" instead, that's really what this is

arrocke (changes requested):
This is looking pretty good! I'm seeing three issues yet and then this can merge:
* Focusing the word shifts everything around slightly
* Button doesn't appear if machine gloss is shown
* Clicking the button should return focus to the input

Pertempto:
It should be fixed now.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -137,12 +137,19 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
               dir={targetLanguage?.textDirection ?? TextDirection.LTR}
             >
               <div className="group-focus-within/input-row:block hidden">
-                {status === 'saved' && (
+                {glossValue && status !== 'approved' && (
```
arrocke: Can we remove this? Right now if you have an empty gloss, and you type something, the button doesn't appear until you move off and come back
```suggestion
                {status !== 'approved' && (
```

arrocke (approved):

packages/web/src/features/translation/TranslateWord.tsx
Pertempto: Then the approval button shows up for a completely empty input, which can't be approved.

Pertempto: I added a `currentInputValue` state variable to help here. It seems to work well, but there might be edge cases I didn't catch.

Pertempto: Can you try it out before we merge?

PR #314, opened 2024-01-04
feat: rich text components

### LINKED ISSUE #313: Rich text components
* Use tiptap to create a rich text editor component `RichTextInput` that supports bold, italic, bullet and numbered lists, indent and outdent
* We'll also need another component `RichText` to render rich text without an editor
* Components should work with HTML strings

arrocke:
## What has changed

* `RichTextInput` component - uncontrolled rich text input compatible with react-hook-form
* `RichText` component - renders rich text without the ability to edit
* Improvements to use of hebrew and greek fonts
* I added the rich text input to the profile view for testing, so I'll revert that commit once this is approved

## Connected Issues

closes #313 

## QA Steps

Use the rich text input on the profile view to check that each button in the toolbar works.

## Post-Deployment

- [x] Nothing required


packages/web/src/styles.css
```diff
@@ -3,16 +3,27 @@
...
+
+    Additionally, Hebrew and to a lesser extent Greek, are harder to read at standard font sizes,
+    so we scale them up a bit so that we don't have to set different font sizes for each language with css.
+   */
   @font-face {
     font-family: 'SBL-Hebrew';
     src: url('./assets/fonts/SBL_Hbrw.ttf') format('truetype');
     font-weight: normal;
+    unicode-range: U+0590-05FF;
+    size-adjust: 140%;
```
arrocke, line 18: I put a comment in the file about these to properties, but it is worth calling out here. Presently we are scaling up Hebrew two sizes from the base font and Greek one size. This enhances readability because the glyphs are smaller in the Hebrew and Greek fonts. By adjusting the size here in the font face, we no longer have to do that. We can use the same font sizes as English and trust that the size for Greek and Hebrew is equivalent for readability.

Additionally, by restricting the unicode range of each font to the characters specific to Greek and Hebrew, we can combine the SBL fonts with Noto Sans so that all characters are rendered by the right font. This means we can use the css class `font-mixed` to render content in all three languages. That's great for lexicon entries.

packages/web/src/features/users/UpdateProfileView.tsx
```diff
@@ -63,11 +70,23 @@ export default function UpdateProfileView() {
...
         <ViewTitle>{t('users:update_profile')}</ViewTitle>
         <Form context={formContext} onSubmit={onSubmit}>
+          <div className="mb-2">
+            <FormLabel id="test">TEST</FormLabel>
+            <RichTextInput
+              aria-labelledby="test"
+              {...formContext.register('test', {
+                value: '<p>Hello Worldוָבֹ֔הוּanother</p>',
+              })}
+            />
```
arrocke: After seeing the consequences of embedding the react hook form registration in custom components in the SortableMultiselectInput, I think we should separate that out like this instead.

The RichTextInput component is proof of concept for how we can design our components to work directly with `formContext.register` without needing to use the `Controller` component like we are in the mutliselect components. This way, if you are using react hook form, you can just use register like I am here. If you aren't using react hook form, `RichTextInput` can works like a normal component.

I'll pull this into a separate ticket to work on, but I wanted to see how that could work for a complex component like this one

packages/web/src/shared/components/form/RichTextInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+        }
+      },
+      focus() {
+        editor?.commands.focus();
+      },
+    }));
+
+    return (
+      <div className="border rounded border-slate-400 focus-within:outline focus-within:outline-2 focus-within:outline-blue-600">
+        <input type="hidden" ref={hiddenInput} name={name} />
```
arrocke, line 85: This hidden input means that the RichTextInput can be used uncontrolled. This means we aren't storing the input value in react state (ie useState). So this component syncs the rich text html to this hidden input. This plus the use of `useImperativeHandle` to provide a custom ref means that react-hook-form can treat this component like a standard `input` element rather than a controlled component which would require `<Controller />`

packages/web/src/styles.css
Pertempto: Awesome!

packages/web/src/shared/components/form/RichTextInput.tsx
Pertempto: :+1: I think I'm following.

packages/web/src/features/users/UpdateProfileView.tsx
Pertempto: > I'll pull this into a separate ticket to work on

You mean that you'll rework our other input components in another issue?

Pertempto:
Wow! I like how well this fits in with our existing styling! It seems to function very well in my testing, I didn't find any glitches.


I noticed that I can't tab to focus the buttons, and that I could use the keyboard to bold/italic. Can you add tooltips with the keyboard shortcuts for the other buttons?

arrocke:
@Pertempto Ok, I added tooltips. The list buttons don't have shortcuts. Tiptap does support markdown, so *, _, or + should work to create a list, but I don't really want to put that in the tooltip. We can add help docs for rich text later if we need to.

Pertempto:
> Tiptap does support markdown, so *, _, or + should work to create a list, but I don't really want to put that in the tooltip.

Yeah that doesn't need to be in the tooltip. It's pretty intuitive to use `*` or `-` to create a list, so when a user tries that it will just work.

Pertempto (approved):
I like it a lot!

Issue #304, opened 2023-12-29
feature flag for word comments


arrocke:
* when `comments` flag is on, show the comments tab in the sidebar
* flag can be turned on with the query parameter `_f=comments`
* flag can be turned on with the `FEAT_COMMENTS` env var
* comments tab should only be available for language members and platform admins

PR #316, opened 2024-01-10
feat: add tab group to sidebar

### LINKED ISSUE #303: add tab panel to sidebar
* Sidebar should have a set of tabs underneath the word
* Initial tab will be titled `Lexicon` and show either the BDB or the LSJ

Image

Pertempto:
## What has changed

Added tab group to the sidebar.
## Connected Issues

#303 

## QA Steps

1. Open the sidebar on the translation page
2. Try the different tabs
3. Check that scrolling works correctly for long lexicon entries

## Post-Deployment

- [ ] Nothing required


packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -45,7 +53,10 @@ export const TranslationSidebar = ({
       <div className="flex flex-row gap-4 items-center">
         <button onClick={onClose} type="button">
           <Icon icon="chevron-down" className="block sm:hidden" />
-          <Icon icon="chevron-right" className="hidden sm:block" />
+          <Icon
+            icon="chevron-right"
+            className="hidden sm:block rtl:rotate-180"
```
Pertempto, line 54: I noticed the button was pointed the wrong way for arabic.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -55,25 +66,48 @@ export const TranslationSidebar = ({
...
-            />
-          </div>
-        )}
+      <div className="grow flex flex-col min-h-0">
+        <Tab.Group>
+          <Tab.List className="flex flex-row gap-1 border-b-2 border-slate-800 md:-ms-3 -mx-4 px-3">
+            {tabTitles.map((title) => (
+              <Tab key={title} as={Fragment}>
+                {({ selected }) => (
+                  <button
```
arrocke: Instead of using Fragment and our own button, we can use `ui-selected:tailwind-class` that the tailwind plugin provides to change the classes when the button is selected

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -35,6 +37,12 @@ export const TranslationSidebar = ({
   );
   const lexiconEntry = lexiconResource?.entry ?? '';
   const { t } = useTranslation(['common', 'translate']);
+
+  const tabTitles = [
+    'translate:lexicon',
+    'translate:notes',
+    'translate:comments',
```
arrocke: For now let's leave the comments tab off. We are going to prioritize the notes tab first, so until the comments are ready to go live, we'll leave that tab hidden with a feature flag

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -55,25 +66,48 @@ export const TranslationSidebar = ({
...
-                __html: DOMPurify.sanitize(lexiconEntry),
-              }}
-            />
-          </div>
-        )}
+      <div className="grow flex flex-col min-h-0">
+        <Tab.Group>
+          <Tab.List className="flex flex-row gap-1 border-b-2 border-slate-800 md:-ms-3 -mx-4 px-3">
+            {tabTitles.map((title) => (
+              <Tab key={title} as={Fragment}>
```
arrocke: The design here is too bold. So a couple suggestions to help it match the tone of the rest of the site.
1. Shrink the border to 1 pixel and change the color to `border-slate-400`. That's a little softer and matches the border color of our input elements.
1. Reduce the padding of the tabs to `px-4 py-1` and remove the `flex-1` so the tabs scale to the size of the content.
1. Use the tab design that removes the bottom border of the selected tab. We can use a spacer element between each tab that has a border, which would then allow each tab to make it's bottom border transparent when selected. The spacer after all of the tabs can have `flex-grow` so that it takes up the remaining space.
1. Add focus outline styles to match the TextInput for accessibility

arrocke (changes requested):
Good work here. A couple suggestions below, and then this should be good

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: Thanks for the suggestions, I'll try to follow them.

I was having a hard time matching the figma design so I just made something similar to the navigation bar.

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: Done

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: I assume the feature flag will be added in the future? I just completely took the tab out for now.

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: Styling fixes done.

I pushed two commits with the focus outline. The first uses `border`, but has a bug where the layout shifts because of `border-2`. The second uses `outline`, but the bottom of the outline looks terrible. Let me know how you would proceed.

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: Ahhh #304

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -68,20 +64,20 @@ export const TranslationSidebar = ({
...
-                    }`}
-                  >
-                    {t(title)}
-                  </button>
-                )}
-              </Tab>
+              <>
+                <Tab
+                  key={title}
+                  className="px-4 py-1 rounded-t-lg outline-none border border-slate-400 ui-selected:border-b-transparent focus:outline-blue-600 outline-offset-0 focus:outline focus:outline-2"
```
arrocke: If you remove outline-none, then you don't need set outline-offset-0 either

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -68,20 +64,20 @@ export const TranslationSidebar = ({
       </div>
       <div className="grow flex flex-col min-h-0">
         <Tab.Group>
-          <Tab.List className="flex flex-row gap-1 border-b-2 border-slate-800 md:-ms-3 -mx-4 px-3">
+          <Tab.List className="flex flex-row md:-ms-3 -mx-4 px-3">
```
arrocke: What do you think about removing the px-3 here so that the tab bottom line goes to the edge?

arrocke (approved):
Two minor suggestions and then this can be merged

packages/web/src/features/translation/TranslationSidebar.tsx
arrocke: I think it is fine for now. We could switch to `focus-visible` which would only show that outline if you are using the keyboard to navigate.

PR #319, opened 2024-01-13
refactor: simplify form abstractions

### LINKED ISSUE #318: Simplify forms
The decision to move react-hook-form registration is creating problems when we want to compose input components into a more complex input like sortable list box. It also makes these components more complex to also be able to support both use inside and outside a form.

Instead we can use the pattern of the RichTextInput component to make all of our inputs compatible with react-hook-form's register function and move the validation configuration to the view.

arrocke:
## What has changed

The goal here is to simplify the abstractions we are using for building forms. Here is what that entails:
* Remove the implicit dependency of form inputs on the FormContext - this allows components to be composed together into a more complex form input that itself is connected to the FormContext by a Controller, and for components to be used outside of a form
* Use Controller for complex uncontrolled inputs - I was trying to hide this inside the input components but it makes things harder to debug. Instead we should just get comfortable without how react-hook-form is designed to be used.
* Use the form context as the source of truth when form state needs to be rendered - Inside a form we don't need to use onChange and useState to keep track of data that we want to render in the template. Instead we can use formContext.watch to get the reactive state

## Connected Issues

closes #318 

## QA Steps

Click on a few forms to ensure they still work as expected.

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/languages/ImportLanguageGlossesView.tsx
+++ b/packages/web/src/features/languages/ImportLanguageGlossesView.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useRef, useState } from 'react';
-import { useForm } from 'react-hook-form';
+import { Controller, useForm } from 'react-hook-form';
 import { Trans, useTranslation } from 'react-i18next';
 import { LoaderFunctionArgs, useLoaderData } from 'react-router-dom';
 import apiClient from '../../shared/apiClient';
@@ -185,17 +185,22 @@ export default function ImportLanguageGlossesView() {
                         <FormLabel htmlFor="import">
                           {t('languages:import_language').toUpperCase()}
                         </FormLabel>
-                        <ComboboxInput
-                          id="import"
+                        <Controller
                           name="import"
-                          className="w-full"
-                          autoComplete="off"
-                          required
-                          aria-describedby="import-error"
-                          items={importLanguages.data.map((language) => ({
-                            label: language,
-                            value: language,
-                          }))}
+                          rules={{ required: true }}
+                          render={({ field }) => (
+                            <ComboboxInput
+                              {...field}
+                              id="import"
+                              className="w-full"
+                              autoComplete="off"
+                              aria-describedby="import-error"
+                              items={importLanguages.data.map((language) => ({
+                                label: language,
+                                value: language,
+                              }))}
+                            />
+                          )}
                         />
                       </div>
                       <div>
--- a/packages/web/src/features/languages/InviteLanguageMemberView.tsx
+++ b/packages/web/src/features/languages/InviteLanguageMemberView.tsx
@@ -63,12 +63,13 @@ export default function InviteLanguageMemberView() {
               {t('users:email').toUpperCase()}
             </FormLabel>
             <TextInput
+              {...formContext.register('email', {
+                required: true,
+              })}
               id="email"
               type="email"
-              name="email"
               className="w-full"
               autoComplete="off"
-              required
               aria-describedby="email-error"
             />
             <InputError
--- a/packages/web/src/features/languages/InviteLanguageMemberView.tsx
+++ b/packages/web/src/features/languages/InviteLanguageMemberView.tsx
@@ -8,7 +8,7 @@ import { useTranslation } from 'react-i18next';
 import { ApiClientError } from '@translation/api-client';
 import Form from '../../shared/components/form/Form';
 import InputError from '../../shared/components/form/InputError';
-import { useForm } from 'react-hook-form';
+import { Controller, useForm } from 'react-hook-form';
 import SubmittingIndicator from '../../shared/components/form/SubmittingIndicator';
 import Button from '../../shared/components/actions/Button';
 import { useFlash } from '../../shared/hooks/flash';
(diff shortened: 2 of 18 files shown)


packages/web/src/features/languages/ImportLanguageGlossesView.tsx
```diff
@@ -185,17 +185,22 @@ export default function ImportLanguageGlossesView() {
...
+                              className="w-full"
+                              autoComplete="off"
+                              aria-describedby="import-error"
+                              items={importLanguages.data.map((language) => ({
+                                label: language,
+                                value: language,
+                              }))}
+                            />
+                          )}
                         />
```
arrocke, line 204: Here is what it looks like to use Controller with a complex input. Before this was all hidden inside the ComboboxInput, but now the connection is made explicit.

packages/web/src/features/languages/InviteLanguageMemberView.tsx
```diff
@@ -63,12 +67,13 @@ export default function InviteLanguageMemberView() {
               {t('users:email').toUpperCase()}
             </FormLabel>
             <TextInput
+              {...formContext.register('email', {
+                required: true,
+              })}
               id="email"
```
arrocke, line 73: For HTML inputs and simple components that wrap HTML inputs, we can just use register

packages/web/src/features/languages/ManageLanguageView.tsx
```diff
@@ -138,7 +138,14 @@ export default function ManageLanguageView() {
...
 
-  const formContext = useForm<FormData>();
+  const formContext = useForm<FormData>({
+    defaultValues: {
+      name: language.data.name,
+      font: language.data.font,
+      textDirection: language.data.textDirection,
+      bibleTranslationIds: language.data.bibleTranslationIds,
+    },
+  });
```
arrocke, line 148: We can specify default values in one place using `useForm`. React hook form will ensure that each input is set correctly from this

packages/web/src/shared/components/form/ButtonSelectorInput.tsx
```diff
@@ -3,33 +3,46 @@ import { useFormContext } from 'react-hook-form';
...
-  required,
   ...props
 }: ButtonSelectorInputProps) {
   const formContext = useFormContext();
-  const hasErrors = !!formContext?.formState.errors[name];
+  const hasErrors = !!(
+    formContext &&
+    name &&
+    formContext.getFieldState(name).error
+  );
```
arrocke: Inside an input component, we can still use the form context to get the error state in order to render the red border because this is filtered by whether the name of the input (if there is one) is in the form or not.

packages/web/src/shared/components/form/ComboboxInput.tsx
```diff
@@ -6,97 +6,36 @@ import {
...
-  isolate?: boolean;
-};
-
-export default function ComboboxInput(props: ComboboxInputProps) {
-  const context = useFormContext();
-
-  if (context && !props.isolate) {
-    return (
-      <Controller
-        control={context.control}
```
arrocke, line 32: We also don't need this wrapper component that decides whether or not to use the Controller or not. That is what got us into trouble with the sortable list input because the presence of the form context caused the input to register itself when we didn't want it to

packages/web/src/shared/components/form/ComboboxInput.tsx
```diff
@@ -6,97 +6,36 @@ import {
...
-  name: string;
+  name?: string;
   items: ComboboxItem[];
   value?: string;
   defaultValue?: string;
-  hasErrors?: boolean;
   up?: boolean;
   onBlur?(): void;
   onChange?(value: string): void;
-  onCreate?(text?: string): void;
```
arrocke, line 88: I also got rid of the onCreate event on the combobox since we aren't using it anymore

packages/web/src/shared/components/form/ComboboxInput.tsx
```diff
@@ -111,6 +50,13 @@ const BaseComboboxInput = forwardRef<HTMLInputElement, BaseComboboxInputProps>(
     const [normalizedInputValue, setNormalizedInputValue] = useState('');
     const [filteredItems, setFilteredItems] = useState<ComboboxItem[]>(items);
 
+    const formContext = useFormContext();
+    const hasErrors = !!(
+      formContext &&
+      name &&
+      formContext.getFieldState(name).error
+    );
```
Pertempto: We're using this pattern a couple places. Could we put it in a helper function that accepts `formContext` and `name` as parameters?

packages/web/src/shared/components/form/ComboboxInput.tsx
```diff
@@ -120,31 +66,11 @@ const BaseComboboxInput = forwardRef<HTMLInputElement, BaseComboboxInputProps>(
             ignoreDiacritics(normalizedInputValue.toLowerCase())
           )
         );
-        const noExactMatch = filteredItems.every(
-          (item) => item.label.normalize('NFD') !== normalizedInputValue
-        );
```
Pertempto, line 125: We're fine to get rid of all this extra logic?

Pertempto (approved):
I'm always a big fan of a code simplification! This looks great, and it seems to work well too.

packages/web/src/shared/components/form/ComboboxInput.tsx
arrocke: Yeah, we don't need to create new elements with a combobox anymore

packages/web/src/shared/components/form/ComboboxInput.tsx
arrocke: This can be simplified to the following, which I don't think is worth abstracting into it's own hook
```
const hasErrors = !!(name && formContext.getFieldState(name).error);
```

Issue #311, opened 2023-12-29
footnotes for words


arrocke:
* sidebar has notes tab
* notes tab has rich text editor for a footnote if the current user is a language translator, otherwise it is readonly
* note is saved to database with edited timestamp and author
* note shows last timestamp and author for language members and platform admin only only

PR #321, opened 2024-01-23
feat: translator notes on words

### LINKED ISSUE #310: translator notes on words
* sidebar has notes tab
* notes tab has rich text editor for translator notes if the current user is a language translator, otherwise it is hidden
* note is saved to database with edited timestamp and author
* note shows last timestamp and author

Pertempto:
## What has changed

- Filled in the translator notes tab of the translation sidebar.
- Added `PATCH /languages/{code}/words/{wordId}/note` for updating the translator note for a word
- Added `GET /languages/{code}/verses/{verseId}/notes` for getting the translator notes for a verse
- Renamed `GET /languages/{code}/verses/{verseId}` to `GET /languages/{code}/verses/{verseId}/glosses`
    - Funny enough, this is how it was already listed in `spec/index.yml`
    - I also updated the spec file for this route, it was quite a bit out of date.

## Connected Issues

#310

## QA Steps

1. Open the sidebar for a word
2. Go to the notes tab
3. Click "Edit"
4. Enter some notes
5. Click "Save"
6. Refresh the page
7. Go to the notes tab for the same word
    - The previously saved note should persist
    - The author and last updated fields should be accurate

## Post-Deployment

- [x] Nothing required


Pertempto:
# To Do
- [x] Get note save working
- [x] Get note load working
- [x] Consolidate migrations
- [x] Update API docs
- [x] Add auto-save
- [x] Update translation terms
- [x] Fix SQL query

packages/api/pages/api/languages/[code]/words/[wordId]/note.ts
```diff
@@ -0,0 +1,52 @@
...
+      subjectId: req.query.code,
+    })),
+    async handler(req, res) {
+      const language = await client.language.findUnique({
+        where: {
+          code: req.query.code,
+        },
+      });
+
+      if (!language || !req.session || !req.session.user) {
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/words/[wordId]/note.ts]: Can we throw an error if the session user isn't defined because that is an error that shouldn't happen. We would want that to return a 500 status code to indicate an error that shouldn't have happened.

packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts
```diff
@@ -0,0 +1,57 @@
...
+            },
+          },
+        },
+      });
+
+      if (verse) {
+        const response: GetVerseTranslatorNotesResponseBody = { data: {} };
+        for (const word of verse.words) {
+          const note = await client.translatorNote.findUnique({
+            where: {
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts]: Just like I did for loading verse glosses, can we write a custom SQL query that loads these notes all in one go rather than using a for loop? That will be far more performant if we only have one trip to the database rather than one for each word in the verse.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -104,7 +141,50 @@ export const TranslationSidebar = ({
                 </div>
               )}
             </Tab.Panel>
-            <Tab.Panel>{t('common:coming_soon')}</Tab.Panel>
+            <Tab.Panel>
+              {/* TODO: add author and timestamp */}
+              <div className="flex flex-col gap-2 pb-2">
```
arrocke: A couple high level UI design thoughts here for discussion. These are my opinions, so I'm open to feedback.

* I think we can condense the design a little bit and communicate the same information. Something like this:

**Translator Notes**
*Updated at 1/2/24 12:30 pm by Adrian Rocke*

* When we want to permit editing, I would prefer to just show the rich text editor and automatically save changes after debouncing. This prevents changes from being lost because somebody didn't press save, and reduces the needed button presses. This works because the rich text input is WYSIWYG where editing and preview are the same thing.
* Finally, we also need to disable editing if the user isn't a language translator. If that's the case, then we should just show the note rather than a rich text editor.

arrocke (changes requested):

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: This makes sense 👍🏽

packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts]: I will try 👍🏽

packages/api/pages/api/languages/[code]/words/[wordId]/note.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/words/[wordId]/note.ts]: Wouldn't that be a 401 error?

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: I will experiment with this later on in the week.

Pertempto:
The notes should be specific to each word ID, correct? My new SQL query has a bug where the same greek word two times in the same verse get the same note.

packages/api/pages/api/languages/[code]/words/[wordId]/note.ts
Comment (arrocke) [reply packages/api/pages/api/languages/[code]/words/[wordId]/note.ts]: Yes technically, but this code shouldn't be reachable as long as authorize guarantees the presence of a user

arrocke:
Yes, each word will have its own note

Pertempto:
@arrocke I have an annoying bug that occurs when I call `translatorNotesQuery.refetch` in `saveNote`. I call `refetch` there so that I can refresh the author name and timestamp string. The annoying thing is that `originalNoteContent` gets updated with the new content.

Try these steps
1. Input a note that ends with a space
2. Pause with the cursor in the note textbox for more than 0.5 seconds


The trailing space gets removed, because apparently the textbox decides to trim it off when sending out the `onChange` signal. This makes for a very annoying user experience.

I'm trying to figure out how I can avoid updating `originalNoteContent` when the debounce auto-save gets called.


I could just get rid of the debounce and only autosave when the user blurs the input box. That functionality seems to work great.

Pertempto:
Maybe we could just figure out how to remove the trimming feature from the `RichTextInput`

Pertempto:
> The notes should be specific to each word ID, correct? My new SQL query has a bug where the same greek word two times in the same verse get the same note.

Fixed

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -35,13 +41,64 @@ export const TranslationSidebar = ({
...
+
+  useEffect(() => {
+    setNoteAuthorName(translatorNote?.authorName ?? '');
+    setNoteTimestampString(
+      translatorNote?.timestamp
+        ? new Date(translatorNote?.timestamp).toLocaleString()
+        : ''
+    );
+    setNoteContent(originalNoteContent);
+  }, [translatorNote, originalNoteContent]);
```
arrocke: Let's try modifying this effect in a few ways:
1. author name and timestamp don't need to be a separate state. They can just be computed like normal: `const noteAuthorName = translatorNote?.authorName ?? ''`. When you can, derive state, don't sync it.
2. However, for the note content, I think we do need to sync state. We can keep a ref of the current word. And only update the note content if that ref indicates that the word has changed. That means we need to replace the rich text editor with that word's note.

I think with these two changes, the react state will always reflect the latest thing that the user has typed, but the author and timestamp will be derived from the latest server state.

Pertempto: Ok, I think I implemented this like you describe. It seems to work now! I had to add the `key` in https://github.com/arrocke/gloss-translation/pull/321/commits/52e3c8bdf185897d5dfb2e53fbe65bbec35cdd7a to resolve a bug where clicking on a word that didn't have any note didn't correctly clear out the rich text input.

packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts
```diff
@@ -0,0 +1,59 @@
...
+          JOIN "TranslatorNote" ON "Word"."id" = "TranslatorNote"."wordId"
+            AND "TranslatorNote"."languageId" = ${language.id}::uuid
+        )
+        --- Now we add in the author's name and combine it all together.
+        SELECT
+          "VerseWord"."id" as "wordId",
+          COALESCE("User"."name", '') AS "authorName",
+          "WordNote"."timestamp",
+          COALESCE("WordNote"."content", '') AS "content"
+        FROM "VerseWord"
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts]: We can simplify this significantly because it is much simpler use case than the glosses. It can just be a single SELECT with a LEFT JOIN for the note and a LEFT JOIN for author.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -35,13 +40,63 @@ export const TranslationSidebar = ({
...
+
+  const canEditNote = userCan('translate', {
+    type: 'Language',
+    id: language,
+  });
+
+  // Only update the content shown in the text editor when the word ID changes.
+  const [originalNoteContent, setOriginalNoteContent] = useState('');
+  const [noteContent, setNoteContent] = useState('');
+  const wordId = useRef('');
```
arrocke: We should just need one note content state rather than two.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -35,13 +40,63 @@ export const TranslationSidebar = ({
...
+  const [needsSaved, setNeedsSaved] = useState(false);
+  // Save the note content after a 500 ms debounce.
+  useEffect(() => {
+    const timeoutId = setTimeout(() => {
+      if (needsSaved) {
+        saveNote(noteContent);
+      }
+    }, 500);
+    return () => clearTimeout(timeoutId);
+  }, [needsSaved, noteContent, saveNote]);
```
arrocke: We should use the [lodash throttle](https://lodash.com/docs/4.17.15#throttle) function in the useCallback. The idea is that the change event can continuously call the saveNote function, but throttle will prevent it from executing too frequently. We actually need throttle and not debounce because debounce won't execute until you stop typing for a bit. We want to save every 15 seconds or so. Using throttle, means that we don't need the useEffect or needsSaved state.
```ts
const saveNote = useCallback(
  throttle(
    () => { /* save note to server */ },
    15000,
    { leading: false, trailing: true }
  ),
  []
)
```

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -104,7 +159,34 @@ export const TranslationSidebar = ({
                 </div>
               )}
             </Tab.Panel>
-            <Tab.Panel>{t('common:coming_soon')}</Tab.Panel>
+            <Tab.Panel>
+              <div className="flex flex-col gap-2 pb-2">
+                {translatorNote?.authorName && (
```
arrocke: Let's add bold text here with the text `Translator's Notes`. We'll need that to distinguish between this input and the Footnote in #311.

arrocke (changes requested):

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: I tried that, it breaks with the debounce logic whenever `setNoteContent` is called, because `noteContent` will get set to something that trims off the spaces, like I mentioned before.

Pertempto: If we could get rid of the debounce logic, it would really simplify the code.

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: This doesn't work quite right. If the note doesn't save instantly when the user blurs the input, they are left questioning if their changes are actually saved correctly. I'll tinker

arrocke: You can use `saveNote.flush()` in the blur handler. Lodash's throttle adds that method to immediately invoke the function.

packages/web/src/features/translation/TranslationSidebar.tsx
arrocke: I push up #324 with an attempt at what I was talking about. This stuff is tricky to get right

packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts]: done

Pertempto:
It seems to be working now, thanks for the help!

package.json
```diff
@@ -18,6 +18,7 @@
     "@prisma/client": "4.11.0",
     "@swc/helpers": "0.5.0",
     "fuzzysort": "2.0.4",
+    "lodash": "^4.17.21",
```
arrocke: Let's lock these versions down

packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts
```diff
@@ -18,29 +18,18 @@ export default createRoute<{ code: string; verseId: string }>()
...
-          "WordNote"."timestamp",
-          COALESCE("WordNote"."content", '') AS "content"
-        FROM "VerseWord"
-        LEFT OUTER JOIN "WordNote" ON "VerseWord"."id" = "WordNote"."id"
-        LEFT OUTER JOIN "User" ON "WordNote"."authorId" = "User"."id"
-        ORDER BY "VerseWord"."id" ASC
+          "TranslatorNote"."timestamp",
+          COALESCE("TranslatorNote"."content", '') AS "content"
+        FROM "Verse"
+        JOIN "Word" ON "Verse"."id" = "Word"."verseId"
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts]: Since we need no fields in the verse table and the word table already has the verse id, we can remove one join by selecting directly from the word table

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -104,7 +150,37 @@ export const TranslationSidebar = ({
...
+                      saveNote(e.target.value);
+                      saveNote.flush();
+                    }}
+                    onChange={async (e) => {
+                      saveNote(e.target.value);
+                    }}
+                  />
+                ) : (
+                  <RichText content={noteContent} />
+                )}
```
arrocke: One thing I had forgotten about is that translator's notes aren't accessible if you aren't a member of the language or a platform admin. So we should only show the header, timestamp and note if you have read access, and then if you have additional edit access (because you are a translator), you get the rich text input.

arrocke (approved):
I had forgotten about one thing. Good to merge after these things are addressed

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: done

packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/notes.ts]: done

package.json
Pertempto: done

PR #324, opened 2024-02-03
refactor: another throttle attempt

arrocke:
Here is an attempt at what I was talking about with throttle. If you want to use this great, or if you want to make further tweaks feel free.

### DIFF
--- a/packages/web/src/features/translation/TranslationSidebar.tsx
+++ b/packages/web/src/features/translation/TranslationSidebar.tsx
@@ -3,7 +3,7 @@ import { useQuery } from '@tanstack/react-query';
 import { Verse } from '@translation/api-types';
 import DOMPurify from 'dompurify';
 import { throttle } from 'lodash';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useAccessControl } from '../../shared/accessControl';
 import apiClient from '../../shared/apiClient';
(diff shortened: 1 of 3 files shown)


packages/web/src/shared/components/form/RichTextInput.tsx
```diff
@@ -66,7 +66,9 @@ const RichTextInput = forwardRef<RichTextInputRef, RichTextInputProps>(
     });
 
     useEffect(() => {
-      editor?.commands.setContent(value ?? '', false);
+      editor?.commands.setContent(value ?? '', false, {
+        preserveWhitespace: 'full',
```
arrocke, line 70: This was the issue with the whitespace. By default tiptap and prosemirror let HTML condense the whitespace like normal

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -62,32 +62,32 @@ export const TranslationSidebar = ({
...
-  }, [language, translatorNotesQuery, word.id, noteContent]);
-
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const debouncedSaveNote = useCallback(
-    throttle(saveNote, 15000, { leading: false, trailing: true }),
-    [language, translatorNotesQuery, word.id, noteContent]
+  const saveNote = useMemo(
+    () =>
+      throttle(
+        async (noteContent: string) => {
```
arrocke, line 77: This idea here is that we have a single function `saveNote` that takes the note content and executes every 15 seconds with the latest note content.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -166,14 +166,13 @@ export const TranslationSidebar = ({
...
-                      setNoteContent(e.target.value);
-                      saveNote();
+                      saveNote(e.target.value);
+                      saveNote.flush();
                     }}
                     onChange={async (e) => {
-                      setNoteContent(e.target.value);
-                      debouncedSaveNote();
+                      saveNote(e.target.value);
                     }}
```
arrocke, line 176: Now we don't have to keep track of any other state, we just have the `noteContent` which changes when the selected word changes and resets the rich text input. But changes just get propagated to the server without having to be synced with react state. The rich text input owns the current state rather than the sidebar.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -166,14 +166,13 @@ export const TranslationSidebar = ({
...
                     key={word.id}
                     name="noteContent"
-                    value={originalNoteContent}
+                    value={noteContent}
                     onBlur={async (e) => {
-                      setNoteContent(e.target.value);
-                      saveNote();
+                      saveNote(e.target.value);
+                      saveNote.flush();
                     }}
```
arrocke, line 173: On blur we can call `saveNote` again, but now we can also use `flush` to force it to execute immediately

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -62,32 +62,32 @@ export const TranslationSidebar = ({
     id: language,
   });
 
-  // Only update the content shown in the text editor when the word ID changes.
-  const [originalNoteContent, setOriginalNoteContent] = useState('');
   const [noteContent, setNoteContent] = useState('');
   const wordId = useRef('');
   useEffect(() => {
-    if (word.id !== wordId.current) {
+    if (translatorNotesQuery.isSuccess && word.id !== wordId.current) {
```
arrocke, line 68: I also had to add `isSuccess` to this effect check otherwise it would run before the notes had loaded over the network

PR #334, opened 2024-02-09
fix: language permissions bug

### LINKED ISSUE #333: language permissions aren't refreshed when you modify them on yourself
**Describe the bug**
When you add or remove language permissions for yourself, the client doesn't update with your new permissions.

**To Reproduce**
1. Log in as an admin.
2. Add a new language
3. Confirm that you can't translate the language
4. Add yourself as a translator
5. Try to translate the language - confirm your permissions haven't updated
6. Now refresh the page
7. Confirm that your permissions have updated and you can translate the language.

**Expected behavior**
The client should refetch your language permissions when you make a change to yourself.

tycebrown:
## What has changed
- The app instantly recognizes self-made updates in the current user's language roles

## Connected Issues

closes #333

## QA Steps

1. As admin, add a new language
2. Confirm that you can't translate the language
3. Add yourself as a translator - confirm that you can translate in the language
4. Try removing and then adding "Translator" from/to your roles in the language - confirm that these actions appropriately update your translation permissions.
5. Now remove yourself from the language members - confirm that you **cannot** translate in the language

## Post-Deployment

- [ ] Nothing required


packages/web/src/features/users/InviteUserView.tsx
```diff
@@ -23,10 +24,13 @@ export default function InviteUserView() {
   useTitle(t('common:tab_titles.invite_user'));
   const flash = useFlash();
 
+  const queryClient = useQueryClient();
+
   const formContext = useForm<FormData>();
   async function onSubmit({ email }: FormData) {
     try {
       await apiClient.users.invite({ email });
+      queryClient.invalidateQueries(['session']);
```
arrocke: We shouldn't need to refetch the session when you invite a new user because that shouldn't ever affect the current user

packages/web/src/features/languages/InviteLanguageMemberView.tsx
```diff
@@ -42,6 +44,7 @@ export default function InviteLanguageMemberView() {
         email: data.email,
         roles: data.roles,
       });
+      queryClient.invalidateQueries(['session']);
```
arrocke: Does this need to match the other places we are doing this?

packages/web/src/features/languages/ManageLanguageView.tsx
```diff
@@ -74,6 +74,7 @@ function useUpdateLanguageMemberMutation() {
       queryClient.invalidateQueries({
         queryKey: languageMembersQueryKey(code).queryKey,
       });
+      queryClient.invalidateQueries({ queryKey: ['session'] });
```
arrocke: In all of these places, it would be good to confirm the affected user is the current user before we force the session to reload. That prevents an unnecessary network request

packages/web/src/shared/accessControl.tsx
```diff
@@ -127,12 +127,15 @@ export interface UserCanOptions {
...
+  // From https://stackoverflow.com/questions/59467758/passing-array-to-useeffect-dependency-list/59468261#59468261
+  // The following variable is used to update the policy when a user's languages change.
+  const languagesStringified = JSON.stringify(user?.languages);
   const policy = useMemo(() => {
     if (status !== 'loading') {
       return createPolicyFor(user);
     }
-  }, [status, user]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [status, user, languagesStringified]);
```
arrocke: How did you decide that we need to add this to the effect? Whenever the session is reloaded, `user` should be a brand new object which should already trigger this effect

arrocke (changes requested):

packages/web/src/features/users/InviteUserView.tsx
tycebrown: Whoops, I should've removed that. Done.

packages/web/src/features/languages/InviteLanguageMemberView.tsx
tycebrown: Yeah, the other two places are for when the user either updates their role as a language member or removes themself from the language. This line invalidates the session when a new language member is invited (I did a brief sanity check to make sure this wasn't another instance of useless code)

packages/web/src/features/languages/ManageLanguageView.tsx
tycebrown: Ah, makes sense.

packages/web/src/features/languages/InviteLanguageMemberView.tsx
arrocke: Sorry, I was asking more about whether this needs to be `{ queryKey: ['session'] }` like the other lines you added. Not sure if it matters.

tycebrown: Oops again, I don't think it matters, except for consistency. Fixed

packages/web/src/shared/accessControl.tsx
tycebrown: I overthought and undertested this. I was thinking it was possible that the user object might remain the same, but the translation roles might change. I added this line and then the call to `invalidateQueries` in `InviteLanguageMemberView` and came to the conclusion that they were both needed to work. 

Fixed.

arrocke (approved):

PR #339, opened 2024-02-11
feat: format prisma pre-commit

### LINKED ISSUE #338: format prisma in pre-commit hook
It would be nice if the prisma files were formatted on commit

Run `nx run db:prisma format` in the `pre-commit` file. 

We could just run `prisma format` manually or add a VSCode extension. 

**Additional context**

tycebrown:
## What has changed
- `nx run db:prisma format` is run before each commit. The formatting changes applied by this command are then added to the commit. 

## Connected Issues

closes #338

## QA Steps

1. Switch your local branch to this branch (`tyce-338-format-prisma`).
2. Locally, add spacing (tabs, spaces, etc.) to `schema.prisma` so that the file is not properly formatted.
   - If you have a VSCode extension which autoformats prisma files on save, you may need to disable that extension or temporarily change the "Language Mode" of `schema.prisma` from "Prisma" to "Plain Text" to test this PR.
3. Commit your changes.
4. Observe the diff of the commit. Confirm that there are no changes observed (since the formatting has been corrected).

## Post-Deployment

- [ ] Nothing required


.husky/pre-commit
```diff
@@ -2,3 +2,7 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 npx lint-staged --relative
+
+nx prisma db format
+# Add the changes made by prisma format to the commit
+git add ./packages/db/src/schema.prisma
```
arrocke: We should use lint staged for this instead. That will automatically stage the changes and only run if the prisma file actually changed in the commit

.gitignore
```diff
@@ -41,6 +41,9 @@ Thumbs.db
 # Next.js
 .next
 
+# NX Cache
+/.nx/cache
```
arrocke: This should already be covered by `.nx/` below

arrocke (changes requested):

.husky/pre-commit
tycebrown: Done.

.gitignore
tycebrown: Done.

arrocke (approved):
Thanks for taking care of this

PR #342, opened 2024-02-15
feat: highlight current verse references in BDB sidebar

### LINKED ISSUE #330: Highlight current reference in the lexicon
When the current verse appears in the lexicon. We want to highlight it so it is easy to find. We will need modify our BDB import to preserve the attributes that mark the verse reference

tycebrown:
## What has changed
- Importing BDB preserves `<a>` tags in lexicon entries.
- Seed is updated with modified BDB.
- Created `bdb-book-ref-names.ts`, which contains the names for books used by the BDB `<a>` tags.
- In the sidebar lexicon tab (for BDB only), references to the current verse are highlighted with the tailwind class `bg-yellow-300`
- At the current moment, verse references in lexicon entries are not clickable.

## Connected Issues

closes #330 
## QA Steps
1. Run `nx run db:prisma migrate reset`
2. Open the web app and select a word (hebrew).
3. Confirm that all references to the current verse are highlighted
4. Change the verse and perform steps 2 and 3 for the newly selected verse

## Post-Deployment

- [x] The seed has already been updated


packages/db/src/scripts/import-bdb.ts
```diff
@@ -317,7 +317,6 @@ function processEntry(entry: string): string {
     entry
       // For some reason sefaria has empty tags in several entries.
       .replaceAll(/<(\/?)big>?/g, '<$1strong>')
-      .replaceAll(/<\/?a[^>]*>/g, '')
```
arrocke, line 320: Can we strip off hrefs here instead of having to do it on the frontend?

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -114,6 +116,27 @@ export const TranslationSidebar = ({
...
   );
+  const { bookId, chapterNumber, verseNumber } = parseVerseId(verse.id);
+  const isHebrew = bookId < 40;
+  const bdbCurrentVerseRef = `${
+    bdbBookRefNames[bookId - 1]
+  } ${chapterNumber}:${verseNumber}`;
+
+  const lexiconEntryHTML = useMemo(() => {
+    if (isHebrew) {
+      const modifiedLexiconEntryElement = document.createElement('div');
```
arrocke: Rather than having to process this in an HTML element before we insert, we can use `useRef` on the lexicon div, and then `useEffect` when the content changes to query for elements and attach the color class

data/seed.dump
arrocke, line 1: what version of postgres do you have installed? I could import this seed because of a version mismatch

arrocke (changes requested):

data/seed.dump
tycebrown: Version 16. Will reseed with Version 14

tycebrown: Reseeded :thumbsup:

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Ah, that's much nicer! Done.

packages/db/src/scripts/import-bdb.ts
tycebrown: Done.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -122,20 +122,16 @@ export const TranslationSidebar = ({
     bdbBookRefNames[bookId - 1]
   } ${chapterNumber}:${verseNumber}`;
 
-  const lexiconEntryHTML = useMemo(() => {
+  const lexiconEntryRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
     if (isHebrew) {
```
arrocke: I think we can remove this condition. Yes, the LSJ doesn't have verses in its content, but the DOM query will come up empty every time and this will be less complicated.

data/seed.dump
arrocke: Hmm... I still can't run the seed script. What does `pg_dump --version` output?

arrocke (approved):
Good to merge when these two last things are addressed

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Done.

tycebrown:
> Hmm... I still can't run the seed script. What does pg_dump --version output?

Oops, I hadn't updated the path to postgres command line tools. Sorry about that. _Now_ it should work (version is 14)

PR #344, opened 2024-02-20
feat: track gloss history

### LINKED ISSUE #340: track gloss history
* Add a `GlossHistory` model the prisma schema with the following schema:
  * wordId PK, FK
  * languageId PK, FK
  * timestamp PK
  * userId FK nullable
  * state nullable - if present, indicates that the gloss state changed
  * gloss nullable - if present, indicates that the gloss text changed
* Whenever the Gloss table is updated, we'll need to add an additional column to GlossHistory

tycebrown:
## What has changed

- Created the `GlossHistory` table as described in issue #340 
- Whenever a user updates a gloss, a row is added to this table
- Whenever the user imports glosses from the old site, a ~~column (🤣)~~ row  is created for each gloss imported.
## Connected Issues
closes #340

## QA Steps

1. Reset the database to use the updated schema (`nx run db:prisma migrate reset`)
2. Query the contents of the `GlossHistory` table. Confirm that it is empty.
3. Change the gloss text for a word. Confirm that a new column is added to the `GlossHistory` table with the updated gloss text.
4. Then, change the gloss approval state without changing the gloss text. Confirm that a new column in `GlossHistory` is created with the updated approval state and `null` for the gloss text.

## Post-Deployment

- [ ] Nothing required


Pertempto:
I plan to review this in the morning

Pertempto:
Lol

> Whenever a user updates a gloss, a column is added to this table

I think you have a mistake there 😆 

that would be QUITE the implementation

packages/db/src/schema.prisma
```diff
@@ -366,3 +372,27 @@ model Footnote {
 
   @@id([wordId, languageId])
 }
+
+/// Represents a gloss update for a particular language.
```
Pertempto:
```suggestion
/// Represents a gloss update for a particular language.
/// Whenever the Gloss model is updated, we'll need to update this model
/// correspondingly with any new or changed fields.
```

Pertempto: This is for the 
> Whenever the Gloss table is updated, we'll need to add an additional column to GlossHistory

bullet point in the issue description.


Also, you should move the `GlossHistory` model up in the file to be immediately below the `Gloss` model.

packages/db/src/schema.prisma
```diff
@@ -366,3 +372,27 @@ model Footnote {
...
+model GlossHistory {
+  /// The word whose gloss is being updated.
+  word       Word        @relation(fields: [wordId], references: [id])
+  /// The ID of the word whose gloss is being updated.
+  wordId     String
+  /// The language of the updated gloss.
+  language   Language    @relation(fields: [languageId], references: [id])
+  /// The ID of the language of the updated gloss.
+  languageId String      @db.Uuid
+  /// The timestamp that the gloss was updated.
```
Pertempto:
```suggestion
  /// The timestamp when the gloss was updated.
```

packages/db/src/schema.prisma
```diff
@@ -44,6 +44,8 @@ model AuthUser {
   translatorNotes    TranslatorNote[]
   /// The list of footnotes authored by this user.
   footnotes          Footnote[]
+  /// The list of gloss updates this user has performed.
+  glossHistory       GlossHistory[]
```
Pertempto: I would move this before `translatorNotes`. It's more important.

packages/db/src/schema.prisma
```diff
@@ -164,6 +166,8 @@ model Word {
   translatorNotes TranslatorNote[]
   // The list of footnotes in different languages for this word.
   footnotes       Footnote[]
+  /// The gloss updates applied to this word.
+  glossHistory    GlossHistory[]
```
Pertempto: I would move this up above `machineGlosses`

packages/db/src/schema.prisma
```diff
@@ -225,6 +229,8 @@ model Language {
   roles               LanguageMemberRole[]
   /// The pending import job if any.
   importJob           LanguageImportJob?
+  // The gloss history for this language.
+  glossHistory        GlossHistory[]
```
Pertempto: I would move this up above `machineGlosses`

packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts
```diff
@@ -58,6 +66,18 @@ export default createRoute<{ code: string; wordId: string }>()
         },
       });
 
+      await client.glossHistory.create({
+        data: {
+          wordId: req.query.wordId,
+          languageId: language.id,
+          userId: req.session?.user?.id,
+          gloss:
+            originalGloss?.gloss !== fields.gloss ? fields.gloss : undefined,
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts]:
```suggestion
            fields.gloss !== originalGloss?.gloss ? fields.gloss : undefined,
```

"If new value doesn't equal the old value" is a bit more intuitive than "If old value doesn't equal new value". We're checking the new value in comparison to the old value, because the new value is the thing that might have changed.

packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts
```diff
@@ -58,6 +66,18 @@ export default createRoute<{ code: string; wordId: string }>()
...
+          state:
+            originalGloss?.state !== fields.state ? fields.state : undefined,
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts]:
```suggestion
            fields.gloss !== originalGloss?.state ? fields.state : undefined,
```

Pertempto:
@arrocke are prisma migrations something that should be included in the Post Deployment section, or are they ran automatically?

Pertempto (changes requested):
The functionality works great! Just a few code improvement suggestions.

packages/db/src/schema.prisma
tycebrown: Oops, I accidentally resolved this suggestion. I don't think there's any problem, but I'll definitely avoid doing that in the future.

tycebrown: Done.

packages/db/src/schema.prisma
tycebrown: Done.

packages/db/src/schema.prisma
tycebrown: Makes sense :+1: Done.

packages/db/src/schema.prisma
tycebrown: Done.

packages/db/src/schema.prisma
tycebrown: Done.

packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts]: Done.

packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/words/[wordId]/gloss.ts]: Done.

packages/db/src/schema.prisma
```diff
@@ -40,6 +40,8 @@ model AuthUser {
   languageRoles      LanguageMemberRole[]
   /// The list of emails awaiting verification.
   emailVerifications UserEmailVerification[]
+  /// The list of gloss updates this user has performed.
+  glossHistory       GlossHistory[]
```
Pertempto: @arrocke I'm going to bring up one concern with your original design....

Should we really name the model `GlossHistory`? It seems like it would be much more accurate to name it `GlossHistoryEntry`. Then `glossHistory` would be of type `GlossHistoryEntry[]`, which would make much more sense.

Pertempto: Another option - 
If I were designing it, I would probably name it `GlossUpdate` over `GlossHistoryEntry`

Pertempto: I just don't think it makes sense for the individual items to be of type `GlossHistory`. That could cause us confusion in the future. You would expect something called "history" to be an array type, not a scalar type.

packages/db/src/schema.prisma
Pertempto: >  I don't think there's any problem, but I'll definitely avoid doing that in the future.

It will probably happen again, I've accidentally done it many times. All you have to do when that happens is click the "Unresolve conversation" button.

arrocke:
> @arrocke are prisma migrations something that should be included in the Post Deployment section, or are they ran automatically?

These are run automatically on merging

packages/db/src/schema.prisma
arrocke: I'm fine calling it `GlossHistoryEntry`. That makes sense to me

tycebrown: Done.

Pertempto (approved):
This works great for me!

fyi @arrocke I did not test the import process

Pertempto:
If you think it's worth the check I can do all the setup on my local env and test it tomorrow.

arrocke:
@Pertempto @tycebrown I will take a look at this tomorrow

packages/db/src/schema.prisma
```diff
@@ -286,6 +292,30 @@ model Gloss {
...
+  /// The timestamp when the gloss was updated.
+  timestamp  DateTime    @default(now())
+  /// The user who updated the gloss.
+  user       AuthUser?   @relation(fields: [userId], references: [id])
+  /// The ID of the user who updated the gloss
+  userId     String?     @db.Uuid
+  /// The updated text of the gloss. Null indicates no change.
+  gloss      String?
+  /// The updated approval state of the gloss. Null indicate no change.
+  state      GlossState?
```
arrocke, line 320: Let's add a field called source that takes a new enum `GlossSource` that has two variants `USER` and `IMPORT`. That way we can distinguish between whether this history entry came from an import or not

packages/lambda-functions/functions/import-language/index.ts
```diff
@@ -87,6 +87,14 @@ export const lambdaHandler = async (event: SQSEvent) => {
               state: GlossState.APPROVED,
             })),
           });
+          await client.glossHistoryEntry.createMany({
+            data: glossData.map(({ wordId, gloss }) => ({
```
arrocke, line 95: We are going to want to save the user who did the import so that we can retrace that if necessary. We'll probably need to add that to the import job in the database so that it is available in the lambda function

arrocke (changes requested):
Nice work on this! I hadn't thought about adding gloss imports to the history, but that's exactly what we want. There are two things yet related to the imported glosses that I missed in my planning that we should resolve yet and then this should be good

packages/db/src/schema.prisma
tycebrown: Done.

packages/lambda-functions/functions/import-language/index.ts
tycebrown: Done.

packages/api/pages/api/languages/[code]/import.ts
```diff
@@ -32,6 +32,7 @@ export default createRoute<{ code: string }>()
       }
 
       res.ok({
+        userId: job.userId ?? undefined,
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/import.ts]: We don't need to send this data to the frontend

arrocke (approved):
One last thing to clean up and then this can be merged

packages/api/pages/api/languages/[code]/import.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/import.ts]: Done

PR #347, opened 2024-02-23
feat: restyle with brand colors

### LINKED ISSUE #331: restyle app using new brand design
We have colors from Andrew and a logo. Those are documented in the epic: #325 

The goal here is to rebuild the site using the brand

arrocke:
## What has changed

* all views rebuilt to use new branding
* admin views and language management views rebuilt with sidebar menu
* user role can now be set when inviting user to the platform

There will undoubtedly be issues since this is so large, but to prevent ones that would continue to bloat this PR, let's only address critical functionality issues and open issues for the others. I'm going to include in that cleaning things up for mobile screens as well.

## Connected Issues

closes #331 

## QA Steps

This touches pretty much everything, so you'll want to try all of the pages.

## Post-Deployment

- [x] Nothing required


Pertempto:
Something seems off about the sidebar background color - https://github.com/arrocke/gloss-translation/issues/348

Pertempto:
Is the manage button supposed to be lowercase like this?

image

Pertempto:
Overall the new styling looks great! I really like the new layout for the settings pages.

I'm not trying to tear your changes apart when I create issues about how we could improve some small things.

Pertempto:
Wow the create user page looks really nice! 🤩

Pertempto:
## Core Tests

- [x] Create new user
- [x] Create new language
- [x] Update language settings
- [x] Add users to language
- [x] User access to languages
- [x] Enter notes in translation sidebar
- [x] Input gloss
- [x] Approve gloss
- [x] Revoke approval of gloss

@arrocke anything else that would be considered core functionality that I should test?

Pertempto:
I have noticed a bug where when you add a new item to a list, that list does not update until you refresh the page.

It happened when I added a language and when I added a user to a language.

I think this would be important enough functionality to justify fixing it in this PR.

Pertempto:
Something trivial haha... This is really more of a `feat:` than a `refactor:` A true refactor doesn't affect functionality. These changes DO impact functionality in a big way.

arrocke:
> @arrocke anything else that would be considered core functionality that I should test?

I think that's a good list

> Something trivial haha... This is really more of a feat: than a refactor: A true refactor doesn't affect functionality. These changes DO impact functionality in a big way.

I think we can leave it as feat because the user can't really do anything new, they just do it a bit differently now.

> Is the manage button supposed to be lowercase like this?

Fixed

> I have noticed a bug where when you add a new item to a list, that list does not update until you refresh the page.

This is fixed as well

Pertempto:
> I think we can leave it as feat because the user can't really do anything new, they just do it a bit differently now.

Alright, for future reference how should we determining the types? I usually go by [this list](https://github.com/pvdlg/conventional-changelog-metahub?tab=readme-ov-file#commit-types) but you seem to be using different definitions. I guess I had in my brain that a refactor is something that wouldn't be worth a bump in version number because the user notices no differences.

arrocke:
I did some more research on conventional commits and I think I was wrong. Seems like most people use refactor as you described. Users shouldn't see any difference because only the code has changed to produce the same result for users. I changed it to feat.

Pertempto:
I really really like the new settings UI! 🤩  The auto-save is especially nice


image



The one thing I noticed is that the LTR indicator doesn't highlight the currently-selected value. Do you want to fix that here or would you like me to create a separate issue?

packages/web/src/app/router.tsx
```diff
@@ -17,11 +27,22 @@ const router = createBrowserRouter([
...
+        loader: () => authorize('administer', 'User'),
+        children: [...userAdminRoutes, ...languageAdminRoutes],
+      },
+      ...userPageRoutes,
+      ...languagePageRoutes,
       ...translationRoutes,
-      ...languagesRoutes,
     ],
   },
+  {
```
Pertempto, line 41: This is no comment on your code, just me trying to understand the router a little better...

How does this item not have to have the `path` field, but the first item (at line 19) does have a `path` field?

Is it because the first one needs to be able to render `/` AND its children routes, while this item is ONLY for rendering its children routes?

Pertempto (approved):
I really like it! There are definitely things that can be improved in the future, but this is an awesome improvement over what we had. It looks like you have more than enough UI design skills to fill in the gap of not having a dedicated UI designer on the team! 😁 Like I said before, the new settings pages are my favorite part of the UI now. Nice job putting in the work to get all this done!!!

I tested the core functionality and didn't find any real show-stoppers. Please respond to my question about the LTR/RTL setting so I can create a new issue for it if you need me to.

I didn't look at the code very deeply at all, just skimmed over all the files. I figured that was appropriate for this size of change. It would take a ton of time and I'd probably lose my brains trying to review/understand every line at one go. Let me know if you want me to look more deeply at any particular code.

Pertempto:
How do I view the landing page in my local env? I ran `nx serve landing` but I'm not sure what port on `localhost` it is using.

Pertempto:
> How do I view the landing page in my local env? I ran `nx serve landing` but I'm not sure what port on `localhost` it is using.

I opened `index.html` in my browser and that did it. Is that the correct way?

arrocke:
> The one thing I noticed is that the LTR indicator doesn't highlight the currently-selected value.

I'm not seeing this issue. Does it still occur when you reset your data seed?

> How does this item not have to have the path field, but the first item (at line 19) does have a path field?

I think the path `'/'` on the first route is redundant. Both root routes are considered [layout routes](https://reactrouter.com/en/main/route/route#layout-routes) because they render nested routes inside of layout components. The AdminView route works similarly, but it prefixes that path with 'admin'.

Pertempto:
> I'm not seeing this issue. Does it still occur when you reset your data seed?

Yes. I reset the DB and still see the bug. Here is the response I'm getting from the API, which seems good.

image

Pertempto:
Clicking the button send an update to the backend with the new `textDirection`. So the API and logic is working correctly, the UI just isn't displaying the currently selected value.

Pertempto:
I just checked in Firefox, the bug appears for me there too.

Pertempto:
Ahhh, I figured it out! I had forgotten to run `npm i` 🤦🏽 

My currently installed tailwind didn't have the `has-` classes.

It displays the current value now! 👍🏽

PR #366, opened 2024-02-28
feat: next unapproved verse

### LINKED ISSUE #356: Next unapproved verse action
In the translation toolbar, add a button to advance to the next verse with words that aren't approved. We also need the keyboard shortcut Alt + N to preform the same action

Image

tycebrown:
## What has changed
- Created `Next Unapproved` button placed in the toolbar next to the verse reference. Pressing this button advances the user to the next verse with unapproved glosses. 
- Created the api path `GET /api/languages/${language}/verses/${verseId}/next-unapproved-verse`

## Connected Issues

closes #356

## QA Steps
1. From any verse, press the button. Confirm that the verse switched to does not have all word glosses approved. Confirm that all verses between the initial verse and the current verse are approved. 
2. Approve all glosses in the current verse
3. Go to the previous verse and press the `Next Unapproved` button. Confirm that the verse approved in step 2 was skipped.

## Post-Deployment

- [ ] Nothing required


tycebrown:
@arrocke Do we want only language members to be able to see this button?

arrocke:
They need to have the translator role on the language. That's the case for the approve button on each word

Pertempto:
I'll have a busy weekend so I can't review this until Monday, if that's not too long of a wait 

@arrocke

packages/web/src/features/translation/TranslationToolbar.tsx
```diff
@@ -46,11 +48,24 @@ export function TranslationToolbar({
     }
   };
 
+  useEffect(() => {
+    const keydownCallback = (e: globalThis.KeyboardEvent) => {
```
arrocke: This shortcut and click handler should only run if the user has permissions

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -327,6 +330,19 @@ export default function TranslationView() {
         onVerseChange={(verseId) =>
           navigate(`/interlinear/${language}/verses/${verseId}`)
         }
+        navigateToNextUnapprovedVerse={async () => {
```
arrocke: Can we move this work into the toolbar and then call `onVerseChange`? That improves the abstraction of the toolbar

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -327,6 +330,19 @@ export default function TranslationView() {
...
+          const data = await apiClient.verses.findNextUnapprovedVerse(
+            verseId,
+            language
+          );
+          if (data && data.verseId) {
+            loadedFromNextButton.current = true;
```
arrocke: What is `loadFromNextButton` for?

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -327,6 +330,19 @@ export default function TranslationView() {
...
+            navigate(`/interlinear/${language}/verses/${data.verseId}`);
+          } else {
+            // TODO: figure out how to handle the situation where ALL words have been glossed (i.e. data.verseId === undefined)
+            flash.error('No unapproved verses');
```
arrocke: I'm ok with what you have here for now. There won't by many languages that are complete any time soon, so we can solve this later

packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts
```diff
@@ -0,0 +1,40 @@
...
+      });
+      if (!language) {
+        return res.notFound();
+      }
+
+      const [result] = await client.$queryRaw<NextUnapprovedVerseResult>`
+        WITH language_gloss as 
+          (SELECT * FROM "Gloss" 
+            WHERE "Gloss"."languageId" = ${language.id}::uuid)
+        SELECT "Verse"."id" as "nextUnapprovedVerseId" FROM "Verse"
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts]: Since we are just using the id of the verse, we don't need to query and join the Verse table. Just filter and return using "Word"."verseId" instead of "Verse".id. One less join is always a good thing

packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts
```diff
@@ -0,0 +1,40 @@
...
+          JOIN "Word" ON "Word"."verseId" = "Verse"."id"
+          LEFT OUTER JOIN language_gloss ON language_gloss."wordId" = "Word"."id" 
+        WHERE "Verse"."id" > ${req.query.verseId} AND (language_gloss."state" = 'UNAPPROVED' 
+                OR language_gloss."state" IS NULL)
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts]: Two suggestions for the WHERE clause for performance and readability:
* We can use a not equals 'APPROVED' comparison instead of two separate equals for null and unapproved
* We can move the language id filter into this clause and eliminate the subquery for language glosses

arrocke (changes requested):
Looking good! A few things to take care of and then this will be ready

packages/web/src/features/translation/TranslationView.tsx
tycebrown: @arrocke This is so that when the translator navigates to the next unapproved verse, the first word of the verse is focused on. My reasoning was that this would make translation more convenient. Do you want me to remove this?

packages/web/src/features/translation/TranslationToolbar.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslationView.tsx
arrocke: Oh right, I forgot that this was how we did that. This is fine how you have it

packages/web/src/features/translation/TranslationView.tsx
tycebrown: The only problem I can see with that we couldn't set `loadedFromNextButton.current` directly in the `TranslationToolbar`. A couple ways around this:
- Pass `loadedFromNextButton` to `TranslationToolbar` (since it's a reference).
- Add a flag parameter to `onVerseChange` to indicate whether the verse should be considered loaded from the next button.

I know this is a small decision, but what do you think we should do?

packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts]: Done.

Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts]: Done.

packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/next-unapproved-verse.ts]: I couldn't get the first one would work and I believe it is due to some quirks in how null comparison works (see [here](https://learnsql.com/blog/null-comparison-operators/)). 

For the second one, I ended up moving the language ID filter into the `ON` clause for the "Gloss" table. This is because if the language ID filter was in the where clause, only verses with non-NULL unapproved glosses would be returned, since all verse would be joined with English glosses and leave no null entries.

I hope that makes sense.

packages/web/src/features/translation/TranslationView.tsx
arrocke: I think I'm ok leaving `loadedFromNextButton` out completely. The reason we do that is so that you can quickly gloss each word, hit the next button and keep going. I don't think that applies for this button in the same way, so it will be simpler for this component if we just leave it out because we can have a single event.

arrocke (changes requested):
Everything else looks good!

packages/web/src/features/translation/TranslationView.tsx

packages/web/src/features/translation/TranslationToolbar.tsx
```diff
@@ -47,10 +50,38 @@ export function TranslationToolbar({
...
+      verseId,
+      languageCode
+    );
+    if (data && data.verseId) {
+      onVerseChange(data.verseId);
+    } else {
+      // TODO: figure out how to handle the situation where ALL words have been glossed (i.e. data.verseId === undefined)
+      flash.error('No unapproved verses');
+    }
+  }, [onVerseChange, flash, verseId, languageCode]);
```
arrocke: Should we be using just verseId rather than data.verseId in here? Regardless the verseId in the dependency array needs to be the same as the verseId used in the function or the reactive stuff won't work right.

arrocke (changes requested):

packages/web/src/features/translation/TranslationToolbar.tsx
tycebrown: My naming was confusing. `verseId` refers to the current verse. `data.verseId` is the next unapproved verse id (`data` here is of type `GetNextUnapprovedVerse`, which, on a side note, should be named `GetNextUnapprovedVerseResponseBody`). I'll change the names be more clear and consistent :+1:

tycebrown: Done.

arrocke: Oh my bad. I misunderstood how this was working.

PR #367, opened 2024-02-28
fix: small screen styling

### LINKED ISSUE #355: Small Screen Width Bug on Landing Page
**Describe the bug**
- On screen sizes below ~450px width, the nav bar buttons start losing the space between them.
- On screen sizes below ~338px width, the body of the UI gets white spaces around it on the bottom and right sides.

**To Reproduce**

1. Open the landing page on desktop
1. Open chrome dev tools with the device toolbar turned on
2. Start diminishing the screen width.
3. Watch the nav bar UI
4. Watch the body UI

**Expected behavior**
- The nav bar should handle screen widths of less than 450px better
    - The logo should not disappear
    - The buttons should re-layout somehow instead of getting squished together
- The body should not get white spaces around it on screen sizes below 338px

**Screenshots**

There might not be very many but there are still iPhone 5 and original SE owners out there, and other people with devices that have such small screen sizes. We don't want to scare them away from the site. You have to set a lower bound on screen width at some point - I usually aim for making it operate fine at 300px width, though we could aim for 320px instead if you want. 
image

- Device: Laptop
- OS: Windows 11
- Browser: Vivaldi

Pertempto:
## What has changed

- Improved the landing page nav bar on small screens

## Connected Issues

#355 

## QA Steps

Test with any screen width `>= 300px`. 
- The logo should be shown
- The title should be shown
- The interlinear button should be shown

Test with any screen width `>= 768px`
- The resources link should be shown
- The volunteer link should be shown
- The go deeper link should be shown

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/landing/index.html
+++ b/packages/landing/index.html
@@ -52,7 +52,7 @@
     </nav>
     <main class="flex-grow overflow-auto">
       <section id="hero" class="relative">
-        <img src="./assets/hero.png" class="w-full" />
+        <img src="./assets/hero.png" class="w-full min-h-80" />
         <div
           class="absolute top-0 bottom-0 w-full flex flex-col items-center justify-center sm:mb-10 md:mb-20 px-8"
         >
--- a/packages/landing/index.html
+++ b/packages/landing/index.html
@@ -14,29 +14,30 @@
 
   <body class="flex flex-col h-screen text-gray-800">
     <nav
-      class="bg-white flex items-center h-20 border-b border-gray-200 relative flex-shrink-0 px-6 md:px-8"
+      class="bg-white flex items-center h-20 border-b border-gray-200 relative flex-shrink-0 px-4 md:px-8"
     >
       <a href="#hero" class="flex items-center mr-8 lg:mr-12">
-        <img src="./assets/logo.png" class="w-14 h-14" />
-        <h1 class="font-bold ms-2 hidden sm:text-lg md:text-2xl sm:block">
-          Global Bible Tools
-        </h1>
+        <img
+          src="./assets/logo.png"
+          class="w-8 h-8 sm:w-10 sm:h-10 md:w-14 md:h-14"
+        />
+        <h1 class="font-bold ms-2 text-lg lg:text-2xl">Global Bible Tools</h1>
       </a>
       <div class="flex-grow md:flex-grow-0"></div>
       <a
-        class="h-full w-24 text-center block pt-[28px] md:pt-[30px] font-bold md:mx-2"
+        class="h-full w-24 text-center hidden md:block pt-[30px] font-bold mx-2"
         href="#resources"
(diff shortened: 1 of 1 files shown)


packages/landing/index.html
```diff
@@ -51,7 +52,7 @@ <h1 class="font-bold ms-2 hidden sm:text-lg md:text-2xl sm:block">
     </nav>
     <main class="flex-grow overflow-auto">
       <section id="hero" class="relative">
-        <img src="./assets/hero.png" class="w-full" />
+        <img src="./assets/hero.png" class="w-full min-h-80" />
```
Pertempto: Before

image


After 

image


This means the image gets slightly stretched vertically on small screens, but I think it it worth it for the text spacing.

packages/landing/index.html
```diff
@@ -14,29 +14,30 @@
...
       <a href="#hero" class="flex items-center mr-8 lg:mr-12">
-        <img src="./assets/logo.png" class="w-14 h-14" />
-        <h1 class="font-bold ms-2 hidden sm:text-lg md:text-2xl sm:block">
-          Global Bible Tools
-        </h1>
+        <img
+          src="./assets/logo.png"
+          class="w-8 h-8 sm:w-10 sm:h-10 md:w-14 md:h-14"
+        />
+        <h1 class="font-bold ms-2 text-lg lg:text-2xl">Global Bible Tools</h1>
```
arrocke: Can we use `text-base sm:text-lg lg:text-2xl` so this doesn't wrap on smaller phones?

packages/landing/index.html
```diff
@@ -14,29 +14,30 @@
...
       </a>
       <div class="flex-grow md:flex-grow-0"></div>
       <a
-        class="h-full w-24 text-center block pt-[28px] md:pt-[30px] font-bold md:mx-2"
+        class="h-full w-24 text-center hidden md:block pt-[30px] font-bold mx-2"
```
arrocke: Can we remove the width and change `mx-2` to `mx-3`? That will make more room on smaller screens for the Go Deeper link I just added.

packages/landing/index.html
arrocke: What if instead of compressing the image, we shrink the spacing between the text and the button so it fits better over the image? We can `mb-8 sm:mb-16` on the paragraph.

arrocke (approved):
Three suggestions and then this should be good

packages/landing/index.html
Pertempto: Done

packages/landing/index.html
Pertempto: Done

packages/landing/index.html
Pertempto: Done

PR #368, opened 2024-02-29
build: cachebust landing page assets when contents change

### LINKED ISSUE #364: cache bust styles output in landing site
The landing site's css is cached in the browser. We need to cache bust this somehow when it's contents change

arrocke:
## What has changed

After copying files to dist folder, renames file with hash based on file contents and then updates assets urls in the index.html. This ensures the hash only changes when the file has actually changed. Pretty crude, but gets the job done.

This probably won't work on windows because it requires the sha256sum or shasum binary. I'm ok with that because this only needs to run on CI, not locally. We don't need file hashes in our dev environments.

## Connected Issues

closes #364

### DIFF
--- a/packages/landing/cachebust
+++ b/packages/landing/cachebust
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+SHA_BIN="sha256sum"
+
+if [ $(uname) = "Darwin" ]; then
+  SHA_BIN="shasum"
+fi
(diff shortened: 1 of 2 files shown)


Pertempto (approved):
Yeah it doesn't work on Windows so I can't test. Makes sense anyways.

PR #370, opened 2024-03-05
feat: admin link goes directly to languages section

### LINKED ISSUE #350: Admin Select Section by Default
When I go to the admin page through the link in the navigation bar, I am presented with an empty page. No section is selected in the sidebar.

By default, select the top section on the admin page

**Additional context**
This depends on the changes from #347

Pertempto:
## What has changed

- Clicking the Admin link in the header goes directly to the languages section, saving the user a click

## Connected Issues

closes #350

## QA Steps

- Click the "Admin" link in the header
- Verify that the browser goes straight to the languages management page

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/app/Header.tsx
+++ b/packages/web/src/app/Header.tsx
@@ -36,7 +36,7 @@ export default function Header() {
         {t('translate:interlinear')}
       </NavLink>
       {userCan('create', 'Language') && (
-        <NavLink to={'/admin'} className={navLinkClasses}>
+        <NavLink to={'/admin/languages'} className={navLinkClasses}>
           {t('common:admin')}
         </NavLink>
       )}
--- a/packages/web/src/app/Header.tsx
+++ b/packages/web/src/app/Header.tsx
@@ -36,7 +36,7 @@ export default function Header() {
         {t('translate:interlinear')}
       </NavLink>
       {userCan('create', 'Language') && (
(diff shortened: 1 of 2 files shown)


packages/web/src/app/Header.tsx
```diff
@@ -36,7 +36,7 @@ export default function Header() {
         {t('translate:interlinear')}
       </NavLink>
       {userCan('create', 'Language') && (
-        <NavLink to={'/admin'} className={navLinkClasses}>
+        <NavLink to={'/admin/languages'} className={navLinkClasses}>
```
Pertempto: I could have implemented this dynamically somehow, so that it automatically goes to the first section of the admin page, no matter what it is. That seemed like unnecessary complexity.

arrocke: We actually need to do this the same way I did with the language settings view because the NavLink renders differently if the URL matches. That won't work for the users section of the admin view because it has a different path. Take a look at the languages/router.tsx file for how I've done this

arrocke (changes requested):

packages/web/src/app/Header.tsx
Pertempto: Ok I attempted something that seems to work with the loader, but I very well could have missed something.

arrocke (approved):

PR #372, opened 2024-03-05
refactor: upgrade lucia and prisma

arrocke:
## What has changed

* Upgraded lucia from v1 to v3 - overall, this simplifies out auth significantly and the code is much easier to understand
  * This upgrade removes the concept of keys - now we just have Users and Sessions
  * Added a table for invitations since keys were handling that before.
  * Added oslo dependency to do password hashing
* Upgraded prisma to v5 - this should bring some performance boosts 

## QA Steps

1. Run `nx run db:prisma migrate dev`
2. Confirm you can still log in with existing accounts
3. Invite a new user to a language
4. Invite a new user to the platform
5. Update email and password for existing user

## Post-Deployment

- [x] Nothing required


packages/api/pages/api/email/verify.ts
```diff
@@ -16,59 +15,38 @@ export default createRoute()
...
-            text: `Your email address for Global Bible Tools was changed to ${verification.email}.`,
-            html: `Your email address for Global Bible Tools was changed to <strong>${verification.email}</strong>.`,
-          });
-        } else {
-          throw new EmailNotVerifiedError(authKey.providerUserId);
-        }
+      if (verification.user.emailStatus === PrismaTypes.EmailStatus.VERIFIED) {
+        await mailer.sendEmail({
+          email: verification.user.email,
+          subject: 'Password Changed',
```
Pertempto: Is this accurate?

Pertempto:
Unrelate to this PR, but I saw this when signing in from an invite.

image

Pertempto:
I updated a user's email, clicked the verification link, then logged back in.

I then attempted to update the user's password

image

```
MissingEmailAddressError: We can't send emails to user with id 018e1324-0c33-e9a5-8642-6a835699f649 because they are missing an email address.
    at Object.sendEmail (webpack-internal:///(api)/./shared/mailer.ts:55:23)
    at Object.handler (webpack-internal:///(api)/./pages/api/users/[userId].ts:88:74)
    at async eval (webpack-internal:///(api)/./shared/Route.ts:171:21)
    at async eval (webpack-internal:///(api)/./shared/Route.ts:230:21)
```

Pertempto:
The bug occurs when I try to update the password for any user.

Pertempto (changes requested):
Everything worked really well except for updating a user's password.

Should the seed file be updated?

arrocke:
I updated the seed file and fixed the mailer

packages/api/pages/api/email/verify.ts
Pertempto: @arrocke Shouldn't this subject be `"Email Changed"`?

Pertempto (approved):
Password update works now :+1:

Please see https://github.com/globalbibletools/gbt/pull/372#discussion_r1517506577, otherwise this looks good!

PR #374, opened 2024-03-11
feat: preview verses in lexicon

### LINKED ISSUE #336: preview verses in lexicons
Verses in lexicon entries should be clickable so translators can see the hebrew and glosses in their language without leaving the verse they are currently working on

Pertempto:
## Connected Issues

closes #336 
## QA Steps

- Either run `nx run db:prisma migrate reset` to reset your DB to the seed file, or run `nx import-lsj db` and `nx import-bdb db` to update the lexicon entries in your DB.
- Go to any verse in the translation screen of the frontend
  - Example: `Matthew 1:3`
- Click on an original language word to open the sidebar
  - Example: `Ἑσρώμ;`
- Click any reference links you see in the sidebar
  - Reference links are in green text
- Verify that the verse preview is loaded correctly for the select reference

NOTE - there may be certain edge cases where the references aren't parsed or handled correctly. Getting the parsing logic correct is tricky. It might be nice to save those remaining edge cases for a separate issue and/or PR, since this one has dragged on so long. I'm pretty sure this PR covers at least 90% of the references correctly.

## Post-Deployment

- [ ] Update lexicons in production DB


Pertempto:
image


@arrocke I've started by attempting to parse out the `ref` tags in the LSJ. How do you want me to handle situations where multiple verses are referred to in one `ref`? For example "Mat.1:6, 12:3". When the user clicks the link should I try to show both verses in the preview? For the first implementation, could I focus on just showing the first verse of the ref in the preview? I don't want to make this PR overcomplicated.


Please ignore the weird styling, it's just to help me debug.

packages/db/src/scripts/import-lsj.ts
```diff
@@ -43,9 +43,8 @@ function toMarkDown(raw: string): string {
       .replaceAll('</Level3>', '')
       .replaceAll('</Level4>', '')
       .replaceAll('<lb />', '')
-      .replaceAll(/<ref=".*">/g, '')
-      .replaceAll(/<ref='.*'>/g, '')
-      .replaceAll('</ref>', '')
+      .replaceAll(/<ref=".*?">(.*?)<\/ref>/g, '<span class="ref">$&</span>')
+      .replaceAll(/<ref='.*?'>(.*?)<\/ref>/g, '<span class="ref">$&</span>')
```
Pertempto: I'm not sure how I'm going to make the jump from this to React portals. I'm sure I can figure it out, but if you have any hints/ideas to jump start me, that would be very helpful!

arrocke:
>  How do you want me to handle situations where multiple verses are referred to in one ref?

Ideally, we would split this into two links, so `Mat.1:6` would be one link and `12:3` a separate link

packages/db/src/scripts/import-lsj.ts
arrocke: My initial thought was something like the following:

1. Use standard DOM methods to insert a div immediately after the verse.
2. Create the portal to the inserted node
3. Add the JSX to the portal

I haven't proven this will work, so as a backup, we can always just create the DOM nodes manually. We may also need to make a separate component so that other state changes don't trigger rerenders, requiring us to reinsert the extra DOM nodes.

Pertempto: Alright, thanks!

Pertempto:
> Ideally, we would split this into two links, so `Mat.1:6` would be one link and `12:3` a separate link

Yeah I agree, but that makes parsing much more complicated. I'm not going to attempt that here. We can always improve in the future.

Pertempto:
I just realized that you can pass a function to `replaceAll`. I'm going to try that, it should make parsing multiple references from one `ref` tag much easier.

Pertempto:
I've got LSJ parsing working and very basic link functionality working

image


I still need to render the correct contents using a portal, and get BDB parsing working.

Pertempto:
Making good progress...

image



Still need to handle rendering Hebrew correctly, and parsing BDB references

Pertempto:
This is ALMOST ready! :tada:

## To Do
- [x] Update seed file
- [x] Update PR description

Pertempto:
@arrocke Can you look at this soon? This one contains a lot of changes, so I'd prefer to get it merged sooner rather than later otherwise I'm going to have to fix a lot of conflicts 😬

arrocke:
I'd like to get to this tomorrow, but realistically it might be Wednesday or Thursday. I have a busy week

Pertempto:
I'll be out of town next week so I won't be able to apply any changes until the fifteenth

packages/web/src/features/translation/VersesPreview.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+function usePreviewQueries(
+  language: string,
+  getContent: boolean,
+  verseIds: string[]
+) {
+  const languagesQuery = useQuery(['languages'], () =>
+    apiClient.languages.findAll()
+  );
+  const translationLanguages = languagesQuery.data?.data ?? [];
```
arrocke: Can we just load the one language rather than all of them?

packages/db/src/scripts/import-bdb.ts
```diff
@@ -319,6 +319,12 @@ function processEntry(entry: string): string {
       .replaceAll(/<(\/?)big>?/g, '<$1strong>')
       // Remove hrefs from hyperlinks
       .replaceAll(/href=".*?"/g, '')
+      // Add ref class so refs can be handled for previews
+      .replaceAll('<a', '<a class="ref"')
```
arrocke: I'm seeing an artifact of the import process. This was in Genesis 1:1 on elohim.
Screen Shot 2024-04-04 at 9 09 15 PM

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -124,14 +130,38 @@ export const TranslationSidebar = ({
...
   useEffect(() => {
     const { current } = lexiconEntryRef;
     // Highlight references to the currently selected verse
     current
       ?.querySelectorAll(`a[data-ref="${bdbCurrentVerseRef}"]`)
       .forEach((element) => element.classList.add('bg-yellow-300'));
-  }, [bdbCurrentVerseRef, lexiconEntry]);
+
+    current?.querySelectorAll<HTMLAnchorElement>(`a.ref`).forEach((element) => {
+      element.onclick = () => {
```
arrocke: We should be able to add a single click event on the wrapper element, and then check the target to see if it an anchor element before executing the click handler logic. Then we don't have to constantly add and remove these events on each individual anchor

packages/web/src/features/translation/verse-utils.ts
```diff
@@ -194,6 +204,43 @@ export function parseReference(reference: string, t: TFunction): string | null {
...
+ * @param reference The string that should be parsed. In the form
+ *                  "bookName chapterNumber:startVerseNumber-endVerseNumber".
+ *                  Example: "Exo 3:14-15". The dash and end verse number are
+ *                  optional.
+ * @param t The i18n translation function to use.
+ * @returns An array of verse IDs, may be empty if no verse IDs can be parsed.
+ */
+export function parseReferenceRange(reference: string, t: TFunction): string[] {
+  const results = reference.split('-');
+  console.log(results);
```
arrocke:
```suggestion
```

packages/web/src/features/translation/VersesPreview.tsx
```diff
@@ -0,0 +1,150 @@
...
+    try {
+      setIsHebrew(isOldTestament(verseIds[0]));
+      setTitle(
+        generateReference(parseVerseId(verseIds[0]), t) +
+          (verseIds.length > 1
+            ? ' - ' +
+              generateReference(parseVerseId(verseIds[verseIds.length - 1]), t)
+            : '')
+      );
+      setIsValid(true);
```
arrocke: All of this is computed state, so ideally, we wouldn't sync this to separate `useState`s, but just compute it inline in the function. If a computed value is expensive to compute, that's what useMemo is for

arrocke (changes requested):
Nice work getting this functional! A few comments to address

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: done

packages/web/src/features/translation/verse-utils.ts
Pertempto: done

packages/web/src/features/translation/VersesPreview.tsx
Pertempto: done

packages/db/src/scripts/import-bdb.ts
Pertempto: working on this...

Pertempto: Done

packages/web/src/features/translation/VersesPreview.tsx
Pertempto: Done.
Just note, I think this causes another HTTP request. The previous query using `languages` as the key I THINK was using the values already loaded by `TranslationView`

Pertempto:
I think all changes have been addressed, I'm going to wait to update seed file until after #382 is merged.

arrocke (approved):
Nice work!

Pertempto:
@arrocke we need to update the lexicons on production with

```
nx import-lsj db
nx import-bdb db
```

arrocke:
I'll do this later today as well as post release notes to slack

PR #375, opened 2024-03-13
feat: approve all glosses

### LINKED ISSUE #357: Approve all glosses button
In the translation toolbar, add a button to approve all glosses in the verse as they are. This should only send one network request rather than one for each word. We also need the keyboard shortcut Alt+A for the same action

Image

tycebrown:
## What has changed

- New `Approve All` button in toolbar and new shortcut `Alt+A` for translators
- New api path: ~~`PATCH /languages/[code]/verses/[verseId]/glosses`~~
  - Updated: `PATCH /languages/[code]/glosses/bulk`

## Connected Issues
closes #357 

## QA Steps
1. Go to a verse. Make sure you are logged out. 
2. Confirm that no  approve all button appears and Alt+A does nothing
3. Log in as admin.
4. Confirm that the button appears.
5. Enter some glosses. 
6. Approve all. Confirm that they are approved by the action
7. Unapprove a gloss. Confirm the button is enabled now.
8. Repeat steps 6 & 7 using Alt+A instead of the button.
## Post-Deployment

- [ ] Nothing required


tycebrown:
TODO:
-----
- [x] Api Docs
- [x] Merge

tycebrown:
@Pertempto or @arrocke Whenever you get time could one of you review this?

arrocke:
@tycebrown Can you resolve merge conflicts? That will make it easier to review. Let me know if you need help

tycebrown:
@arrocke Merged :+1:

Pertempto:
I can review tomorrow morning

packages/web/src/public/locales/es/translate.json
```diff
@@ -1,4 +1,6 @@
 {
+  "all_glosses_approved": "Todos los brillos aprobados",
+  "approve_all": "Aprobar todo",
```
Pertempto: @JCampana98, do these look good?

Pertempto:
The functionality seems to work great! Nicely done!!


The one thing I noticed is that the approve all button flashes to disabled when I insert new glosses and then quicly goes back to enabled. I think this happens because the new gloss is getting sent to the backend. Is there any way to avoid flashing the button like that ? I would imagine it would be distracting to translators as they insert glosses.

packages/api/public/spec/paths/translation/verse-glosses.yml
```diff
@@ -61,3 +61,43 @@ get:
...
+    required: true
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            data:
+              type: object
+              additionalProperties:
+                type: object
```
Pertempto: This `additionalProperties` isn't quite right. See `localhost:4300/explorer` to see how it shows up.

image

I think if you remove these two lines and then dedent the rest of the lines, it should fix it.

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
```diff
@@ -160,4 +167,113 @@ export default createRoute<{ code: string; verseId: string }>()
...
+                Prisma.sql`(${wordId}, ${gloss ?? null}, ${
+                  state ?? null
+                }::"GlossState")`
+            )
+          )}) as gloss_updates("wordId", "gloss", "state")
+          WHERE "Gloss"."languageId" = ${language.id}::uuid
+            AND gloss_updates."wordId" = "Gloss"."wordId"
+        `;
+      }
+      await client.glossHistoryEntry.createMany({
```
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Shouldn't this also be in the `if` block?

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (Pertempto) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts:1]: @arrocke can you double check the SQL logic? It looks pretty good to me, but you are much better at it than I.

packages/web/src/features/translation/TranslationView.tsx
Pertempto, line 1: @arrocke can you double check the mutation logic here? I'm still not very comfortable will the whole mutation system.

Pertempto (changes requested):
It all looks pretty good to me! Great job! 

I just noticed the one thing that needs fixed in the YML file.

Hopefully @arrocke can double check the database and mutation logic.

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
```diff
@@ -160,4 +167,113 @@ export default createRoute<{ code: string; verseId: string }>()
       }
     },
   })
+  .patch<PatchVerseGlossesRequestBody, void>({
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: The way that I want to handle bulk requests is with a post request to `/languages/{code}/glosses/bulk`. There isn't a real need to limit this to a single verse since the payload includes each word being updated.

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
```diff
@@ -160,4 +167,113 @@ export default createRoute<{ code: string; verseId: string }>()
...
+      if (entriesToUpdate.length > 0) {
+        await client.$executeRaw`
+          UPDATE "Gloss" SET 
+            "gloss" = COALESCE(gloss_updates."gloss", "Gloss"."gloss"),
+            "state" = COALESCE(gloss_updates."state", "Gloss"."state")
+          FROM (VALUES ${Prisma.join(
+            entriesToUpdate.map(
+              ({ wordId, gloss, state }) =>
+                Prisma.sql`(${wordId}, ${gloss ?? null}, ${
+                  state ?? null
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Ideally, we would do an INSERT INTO ... ON CONFLICT which would be an "upsert many" query. That would either insert the record if it doesn't exist, or update it if it does. That would reduce the db queries to three - load the data for comparison, upsert glosses, create gloss history entries.

The other major thing is that we need to wrap all of this in a transaction. That's not something we've been good about, but it ensures this data changes together or not at all

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
```diff
@@ -160,4 +167,113 @@ export default createRoute<{ code: string; verseId: string }>()
...
+      });
+      if (!language) {
+        return res.notFound();
+      }
+
+      const glosses = Object.fromEntries(
+        (
+          await client.$queryRaw<PrismaTypes.Gloss[]>`
+        SELECT "wordId", "gloss", "state" FROM "Gloss" JOIN "Word" ON "Gloss"."wordId" = "Word"."id"
+          WHERE "Word"."verseId" = ${req.query.verseId} AND "Gloss"."languageId" = ${language.id}::uuid
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Here we want to load the glosses for the words in the word id keys in the bulk update.

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -312,9 +313,59 @@ export default function TranslationView() {
...
+              { gloss: glossAsDisplayed, state: GlossState.Approved },
+            ])
+        );
+        await apiClient.verses.updateVerseGlosses(verseId, language, {
+          data,
+        });
+      }
+    },
+    onSuccess: async () => {
+      await targetGlossesQuery.refetch();
```
arrocke: The better way to reload state is use `queryClient.invalidateQueries`. That's because there could be multiple places that use the same query key to load the same data from the cache and `invalidateQueries` handles all of them.

arrocke (changes requested):

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Next time could you include this sort of information in the issue description so we don't waste time on the wrong implementation?

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Done.

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Done 👍

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Done :+1:

packages/web/src/public/locales/es/translate.json
JCampana98: Not quite 😅 
```suggestion
  "all_glosses_approved": "Todas las glosas están aprobadas",
  "approve_all": "Aprobar todo",
```

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Would it be possible to skip using a transaction and instead use `WITH` clauses?

Comment (arrocke) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Why is it better to skip a transaction? We can just do this:
```ts
await client.$transaction(async (tx) => {
  // Use tx instead of client to run the queries, logic, and upserts
})
```

packages/web/src/public/locales/es/translate.json
tycebrown: Done.

packages/api/public/spec/paths/translation/verse-glosses.yml
tycebrown: @Pertempto The only thing is that, at the current moment, data is passed in as a record, with the `wordId` as the key:
```
  data: {
    [wordId: string]: { gloss?: string; state?: GlossState };
  };
```
I think we _could_ fix the YAML file without changing the request type, but do you think it be better to change the request body type and go with the fix you prescribed?

tycebrown:
> The one thing I noticed is that the approve all button flashes to disabled when I insert new glosses and then quicly goes back to enabled. I think this happens because the new gloss is getting sent to the backend.

That was the culprit. Done.

packages/web/src/features/translation/TranslationView.tsx

packages/api/public/spec/paths/translation/verse-glosses.yml
Pertempto: We don't change the request type just to match the documentation, you change the documentation to match the request. I guess I don't have enough experience in swagger to give you a good answer on how you'll make the swagger match your request, but there's plenty of good examples and documentation online.

tycebrown: Done.

packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Done.

Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/glosses.ts]: Done.

packages/api/public/spec/paths/translation/verse-glosses.yml
Pertempto: Nice work, that looks much better! Are those word IDs accurate examples? They look really short.

Pertempto (approved):
All the changes I requested have been resolved. Nice work on this one!

packages/api/public/spec/paths/translation/verse-glosses.yml
tycebrown: Oh shoot, you're right! :rofl: Those were verseIds. I'll fix that.

tycebrown: Fixed :+1:

packages/api/pages/api/languages/[code]/glosses/bulk.ts
```diff
@@ -0,0 +1,101 @@
...
+            INSERT INTO "GlossHistoryEntry"("languageId", "userId", "wordId", "gloss", "state", "source") 
+            VALUES ${Prisma.join(
+              patchedGlosses.map((patchedGloss) => {
+                const oldGloss = oldGlosses[patchedGloss.wordId];
+                return Prisma.sql`
+                    (${language.id}::uuid, 
+                    ${req.session?.user?.id}::uuid, 
+                    ${patchedGloss.wordId}, 
+                    NULLIF(${patchedGloss.gloss}, ${oldGloss?.gloss}), 
+                    NULLIF(${patchedGloss.state}, ${oldGloss?.state})::"GlossState",
```
Comment (arrocke) [inline packages/api/pages/api/languages/[code]/glosses/bulk.ts:90]: We only want to set the gloss and state in the history entry if they changed, otherwise they can be null. If we make that change, then I don't think we need to return anything from the previous query.

arrocke (changes requested):
One minor change to how we are saving gloss history, and then this is good! Nice work!

packages/api/pages/api/languages/[code]/glosses/bulk.ts
Comment (tycebrown) [reply packages/api/pages/api/languages/[code]/glosses/bulk.ts:90]: @arrocke I'm a little confused. Won't `NULLIF` make the `"gloss"` and `"state"` null in the history entry if there was no change between the `patchedGloss` and `oldGloss`? Do you mean that we should remove the `patchedGlosses` variable (which is set to upsert-many result) and use the `entriesToPatch` variable instead?

Comment (arrocke) [reply packages/api/pages/api/languages/[code]/glosses/bulk.ts:90]: Oh woops, that's my bad. I misunderstood what nullif was doing. I was thinking we should be able to insert the history entries without needing to return anything from the first query, but now that I understand nullif, I think this is fine. It's seems fast enough, so let's proceed

arrocke (approved):

PR #380, opened 2024-03-25
feat: reset password

### LINKED ISSUE #373: Reset Password
As far as I am aware, there is currently no way to trigger a password reset from our frontend.

Reset password button on the login page, clicking it would trigger an email with a password reset link. Clicking the link would take them to a page where they could input their new password.

Administrators could trigger password reset emails from the admin page

tycebrown:
## What has changed
- Added a "Forgot Password?" button in the login page.
- Added two new pages:
  - `/forgot-password` - A page the user can email a link to themself which leads to `/reset-password`.
  - `/reset-password?token` - A page where the user submits their new password.
- Added two new api paths:
  - `POST /api/auth/forgot-password` - Creates a `ResetPasswordToken` from an email address passed in the body.
  - `POST /api/auth/reset-password` - Deletes the created `ResetPasswordToken`, updates the user's password, and logs the user in. 
- Created a schema in prisma for `ResetPasswordToken`

## Connected Issues

closes #373 

## QA Steps
1. Go to the Login page.
2. Press the "Forgot Password?" button. 
3. Enter "platform_admin@example.com" into the email. 
4. Press the "Reset Password" button. Confirm that a success message flashes.
5. If you have no email server set up, go to the api log and confirm that a reset password link has been sent. Go to that link.
6. Enter a new password and press "confirm" button.  Confirm that clicking the button logs you in. 
7. Log out. 
8. Confirm you can log back in with the new password.

## Post-Deployment

- [ ] Nothing required


Pertempto:
Your translation terms need cleaned up... you have a couple duplicated between two different files. They all have to do with the same topic, so I would centralize them in one file, probably `users.json`

packages/web/src/public/locales/en/users.json
```diff
@@ -21,15 +21,19 @@
     "user_exists": "A user with email \"{{email}}\" already exists."
   },
   "first_name": "First Name",
+  "forgot_password": "Forgot Password?",
```
Pertempto: Do we really need version with a question mark and one without? Why not combine them?

packages/web/src/features/users/ResetPasswordView.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { ApiClientError } from '@translation/api-client';
+import Form from '../../shared/components/form/Form';
+import { useTranslation } from 'react-i18next';
+import { useNavigate, useSearchParams } from 'react-router-dom';
+import useAuth from '../../shared/hooks/useAuth';
+import useTitle from '../../shared/hooks/useTitle';
+
+export default function ResetPasswordView() {
+  const { t } = useTranslation(['users', 'common']);
+  useTitle(t('common:tab_titles.reset_password'));
```
Pertempto:
```suggestion
  useTitle(t('users:reset_password'));
```

I would remove the question mark from `users:reset_password` and use it here. No need to have duplicate terms.

packages/web/src/public/locales/en/common.json
```diff
@@ -42,6 +43,7 @@
     "login": "Login",
     "manage_language": "Manage {{languageName}}",
     "new_language": "Create Language",
+    "reset_password": "Reset Password",
```
Pertempto: I couldn't find usages for this one? If they exist, replace with `users:reset_password`

packages/web/src/features/users/ResetPasswordView.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <ModalView className="max-w-[480px] w-full">
+      <ModalViewTitle>{t('users:new_password')}</ModalViewTitle>
+      <Form
+        context={formContext}
+        onSubmit={onSubmit}
+        className="max-w-[300px] w-full mx-auto"
+      >
+        <div className="mb-6">
+          <FormLabel htmlFor="newPassword">
```
Pertempto, line 61: I think we should have a confirm password input here, what if the user accidentally mis-types their new password?

Pertempto (changes requested):
The functionality seems to work well. 

@arrocke can you double check the API/DB design for any red flags?

packages/web/src/public/locales/en/users.json
tycebrown: Done.

packages/web/src/features/users/ResetPasswordView.tsx
tycebrown: Done :+1: Question mark removed from `users:forgot_password`

packages/web/src/public/locales/en/common.json
tycebrown: Done.

packages/web/src/features/users/ResetPasswordView.tsx
tycebrown: Done.

packages/web/src/public/locales/es/users.json
Pertempto, line 1: @jcampana98 if you have time, can you verify these new Spanish terms?

Pertempto:
One potential security/privacy issue...

If you enter an email that doesn't exist in the system in the forgot password page, it gives an error. If you enter an email of a user in the system, it gives a success message.

That means we are leaking out which emails are in our system. I think the better practice is to return a success message from the API, whether or not the user exists, and just silently ignore emails that don't exist in the system. This way the emails of our users stay private from bad actors.

@arrocke can you weigh in on this?

arrocke:
Yeah, this is correct. The message should be something like `Please check your email to reset your password" regardless of whether it succeeds or not

packages/api/pages/api/auth/forgot-password.ts
```diff
@@ -0,0 +1,39 @@
...
+      }
+      const { token } = await client.resetPasswordToken.upsert({
+        where: { userId: user.id },
+        update: { expires: Date.now() + 60 * 60 * 1000 },
+        create: {
+          userId: user.id,
+          token: randomBytes(12).toString('hex'),
+          expires: Date.now() + 60 * 60 * 1000,
+        },
+      });
```
arrocke: A user can have multiple tokens, so we can just create a new one each time. This ensures that a token can't be reused after it expires. Otherwise a compromised token could be used by requesting a new reset

packages/api/pages/api/auth/reset-password.ts
```diff
@@ -0,0 +1,44 @@
...
+      if (!user) {
+        throw new InvalidTokenError();
+      }
+
+      await client.$transaction([
+        client.user.update({
+          where: { id: resetPasswordToken.userId },
+          data: { hashedPassword: await scrypt.hash(req.body.password) },
+        }),
+        client.resetPasswordToken.delete({ where: { token: req.body.token } }),
```
arrocke, line 37: After the password is changed we need to send the same email we do when the password is changed from the profile form

packages/web/src/features/users/LoginView.tsx
```diff
@@ -71,9 +72,19 @@ export default function LoginView() {
...
+              {t('users:password').toUpperCase()}
+            </FormLabel>
+            <Button
+              small
+              variant="tertiary"
+              className="italic text-xs"
+              to={'/forgot-password'}
+            >
+              {t('users:forgot_password')}
+            </Button>
```
arrocke: I forgot to include this in the ticket, but I had originally designed the forgot password to be below the login button. That's a pretty common place it goes in login views

Screen Shot 2024-03-27 at 9 23 35 PM

packages/web/src/features/users/ResetPasswordView.tsx
```diff
@@ -0,0 +1,105 @@
...
+import useTitle from '../../shared/hooks/useTitle';
+
+export default function ResetPasswordView() {
+  const { t } = useTranslation(['users', 'common']);
+  useTitle(t('users:reset_password'));
+
+  const { refreshAuth } = useAuth();
+
+  const [search] = useSearchParams();
+  const token = search.get('token') ?? undefined;
```
arrocke: It would be good to have a useEffect that confirms there is a token, and if not shows an error message and redirects to the home page.

packages/web/src/features/users/ResetPasswordView.tsx
```diff
@@ -0,0 +1,105 @@
...
+              {...formContext.register('confirmNewPassword', {
+                validate: {
+                  confirms: (value: unknown) =>
+                    value === formContext.getValues().newPassword,
+                },
+              })}
+              type="password"
+              id="confirm-new-password"
+              className="w-full"
+              autoComplete="new-password"
```
arrocke: This autocomplete should go on the other text input as well.

packages/web/src/features/users/ForgotPasswordView.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const formContext = useForm<{ email: string }>();
+  async function onSubmit({ email }: { email: string }) {
+    try {
+      await apiClient.auth.forgotPassword({ email });
+      flash.success(t('users:reset_password_link_sent', { email }));
+    } catch (error) {
+      if (error instanceof ApiClientError && error.status === 401) {
+        flash.error(t('users:errors.invalid_auth'));
+      } else {
```
arrocke: We can remove this 401 error check now too since that error isn't possible anymore

packages/web/src/features/users/ResetPasswordView.tsx
```diff
@@ -0,0 +1,105 @@
...
+            aria-describedby="password-error"
+          />
+          <InputError
+            id="password-error"
+            name="newPassword"
+            messages={{
+              required: t('users:errors.password_required'),
+            }}
+          />
+          <div className="mb-4">
```
arrocke: This div is nested one level too deep. We also need to adjust the margins to match how margins are created on other pages. mb-4 between inputs and mb-6 on the input before the submit button.

arrocke (changes requested):
A couple things yet to address and then this should be good. Thanks for your work on this.

packages/web/src/public/locales/es/users.json
JCampana98: They are good enough 😄

packages/api/pages/api/auth/forgot-password.ts
tycebrown: Done.

packages/api/pages/api/auth/reset-password.ts
tycebrown: Done.

packages/web/src/features/users/LoginView.tsx
tycebrown: Done. By the way, is there a place where our new figma designs are located? It seems we've long moved away from https://www.figma.com/file/GYKTJWl1otgwpmA9ZLLthE/Gloss-Translation-App?type=design&node-id=1023-276&mode=design&t=pYANiBhuPrIb150t-0

packages/web/src/features/users/ResetPasswordView.tsx
tycebrown: Done. This was a lousy mistake. I will make an effort to be thorough in my testing.

packages/web/src/features/users/ForgotPasswordView.tsx
tycebrown: Done.

Pertempto:
A bug introduced by your new changes:

A response 
```{"errors":[{"code":"AlreadyExists"}]}```
is given if the user clicks the Reset Password button when an existing reset password token exists for that user already.

File: packages/web/src/features/users/ResetPasswordView.tsx (hunk 9 above)
Comment (tycebrown) [reply packages/web/src/features/users/ResetPasswordView.tsx]: Done.

File: packages/web/src/features/users/ResetPasswordView.tsx (hunk 8 above)
Comment (tycebrown) [reply packages/web/src/features/users/ResetPasswordView.tsx]: Done. In development mode, this flashes twice due to the strict mode setting. I don't think this is a problem since we immediately navigate away from `/reset-password` on the error.

**Comment by tycebrown** [comment]
> A response
{"errors":[{"code":"AlreadyExists"}]}
is given if the user clicks the Reset Password button when an existing reset password token exists for that user already.

Ah, I forgot to mention; I needed to update the schema to allow for multiple tokens per user (I had created a flawed schema where the user had a one-to-one relationship with a reset-password token), so you'll need to run `nx prisma db migrate reset`. Sorry for the confusion.

**Comment by tycebrown** [comment]
@JCampana98 If you have time, could you verify the following Spanish term that I added?

English:
```
"no_reset_password_token": "Error, no reset password token provided; link is broken.",
```
Spanish:
```
"no_reset_password_token": "Error, no se proporcionó ningún token para restablecer la contraseña; el enlace está roto.",
```

Pertempto:
Alright it works correctly now!



> > A response
> > {"errors":[{"code":"AlreadyExists"}]}
> > is given if the user clicks the Reset Password button when an existing reset password token exists for that user already.
> 
> Ah, I forgot to mention; I needed to update the schema to allow for multiple tokens per user (I had created a flawed schema where the user had a one-to-one relationship with a reset-password token), so you'll need to run `nx prisma db migrate reset`. Sorry for the confusion.

Pertempto (approved):
Looks great! Nice work!

All the concerns I raised are addressed. @arrocke will still need to check that his were covered.

arrocke (approved):

PR #381, opened 2024-03-29
refactor: implement isOldTestament helper function

Pertempto:
## What has changed

Added `isOldTestament` helper function.
- If we ever change the book ID system, then we will only need to update the condition `bookId < 40` ONE place, instead of many places.
- This is a way to move the magic number to only one place.

## Connected Issues

None. Recently I've been blessed by the idea that refactoring changes should be kept completely separate from new feature and bug fix changes. It helps the review process go much quicker, because the refactoring diffs aren't mixed in with the actual diffs that change functionality. I got the concept from Kent Beck's _Tidy First?_ book.

We've been trying it out at work and it's reduced review times greatly. This is a tiny code improvement I found while working on #374, so I thought I'd hold it off for a separate PR. We probably should merge this after that one to reduce diffs there.

## QA Steps

- Check that hebrew text is still right-aligned in the main translation view
- Check that hebrew text is still right-aligned in the verse previews in the lexicon tab of the sidebar

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/features/translation/verse-utils.ts
+++ b/packages/web/src/features/translation/verse-utils.ts
@@ -36,7 +36,8 @@ export function verseCount(bookId: number, chapterNumber: number): number {
  * @returns true when the verse ID comes from an Old Testament book, false otherwise.
  */
 export function isOldTestament(verseId: string): boolean {
-  return parseVerseId(verseId).bookId < 40;
+  const matthewBookId = 40;
+  return parseVerseId(verseId).bookId < matthewBookId;
 }
 
 /**
--- a/packages/web/src/features/translation/verse-utils.ts
+++ b/packages/web/src/features/translation/verse-utils.ts
@@ -31,6 +31,15 @@ export function verseCount(bookId: number, chapterNumber: number): number {
 }
 
 /**
+ * Check if the verse ID comes from an Old Testament book.
+ * @param verseId The verse ID to check, in the form BBCCCVVV.
+ * @returns true when the verse ID comes from an Old Testament book, false otherwise.
+ */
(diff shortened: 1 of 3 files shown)


packages/web/src/features/translation/verse-utils.ts
```diff
@@ -30,6 +30,15 @@ export function verseCount(bookId: number, chapterNumber: number): number {
   return verseCounts[bookId - 1][chapterNumber - 1];
 }
 
+/**
+ * Check if the verse ID comes from an Old Testament book.
+ * @param verseId The verse ID to check, in the form BBCCCVVV.
+ * @returns true when the verse ID comes from an Old Testament book, false otherwise.
+ */
+export function isOldTestament(verseId: string): boolean {
+  return parseVerseId(verseId).bookId < 40;
```
Pertempto:
```suggestion
  const matthewBookId = 40;
  return parseVerseId(verseId).bookId < matthewBookId;
```

I was going to add an explanation comment, but then I realized a well named variable would make the magic number self-explanatory.

Pertempto:
@arrocke i know we've had some discussion before about code-cleanup stuff slowing down code review. Hopefully splitting it up into a separate pull request helps resolve that problem. Please let me know your thoughts on this strategy. Does it cause additional problems for you?

It's worked wonders for us at work (@jcampana98 can a test to that 😆), but maybe it's not a good fit for the `gbt` project.

arrocke (approved):
I don't have a strong preference. This was easy to review, so as long as you can articulate the value of the change in the PR then these are fine!

Issue #376, opened 2024-03-15
Migrate gloss data model to support phrasal glosses


arrocke:
Some languages require the ability to set a gloss on a group of words, what we will call a phrase. So we need to adapt our model to support multiple words on the same gloss.

The new db structure will be something like the diagram below.
```mermaid
erDiagram
    Language ||--o{ Gloss : has
    Word }o--|{ GlossPhraseWord : connects
    Gloss ||--|{ GlossPhraseWord : connects
```

Unfortunately, this means that all glosses will require an extra join. The Gloss table will need its own integer primary key, so that GlossPhrase can be a join table with a primary key that consists of a foreign key to the Word table and a foreign key to the Gloss table. Now we can associate multiple words with a single gloss.

We will also need to move footnotes and translator notes to point to the gloss table rather than the word. This will allow a phrasal gloss to share these notes regardless of which word is selected. The GlossHistory table will also need to point to the gloss table as a foreign key. We'll need to start recording the word ids that comprise the gloss. Since this is just historical data, we could use an array field to record the list of word for the gloss, that way we can reconstruct the history of which gloss phrase each word is part of.

The goal of this ticket is to introduce this structure into the database while preserving the current logic. We will add the API calls and UI elements for phrasal glosses in a different ticket.

Pertempto:
Just to make sure I'm following...

The gloss phrases will consists of different combinations of glosses, depending on the language? In one language this sequence of three original language words will have three target language glosses, but in another language that sequence of three words might be one phrase with one one gloss?

That sounds tricky to implement in the UI, but I'm guessing you're working on the design for that?

Pertempto:
This is also to clarify my understanding, I'm not critiquing your diagram, just trying to make sure I understand it clearly.

It's been a couple years since I worked with database diagrams... so I might be confused on a few of these. I'm using [this chart](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fi.stack.imgur.com%2F9tvIZ.gif&f=1&nofb=1&ipt=2bf0de7f36f89238b19804e31afc43bfe15854fcc8ac5e691485e751c8fdd27e&ipo=images) for reference.

- One gloss is associated with many languages
    - Isn't this backwards? I thought a gloss represented a word in one particular language. So a language would have many glosses, but a single gloss is in only one language.
- Many words are associated with many gloss phrases
[...]

PR #382, opened 2024-03-31
refactor: migrate db schema to track phrases

arrocke:
## What has changed

* Added phrases to the database
* All writes (glosses and notes) that used to point to a word, now additionally point to a phrase
* A sql script to connect existing glosses and notes to a single word phrase

## Connected Issues

This is the first two steps of #376 as well as the migration script we can use for step three

## QA Steps

1. Run `nx run db:prisma migrate dev`
2. Create a gloss or a note
3. Look in the database to see that a phrase and phrase word have been created and connected to the gloss and phrase.
4. No existing user facing behavior should chnage

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/db/src/schema.prisma
+++ b/packages/db/src/schema.prisma
@@ -310,7 +310,7 @@ model PhraseWord {
   phrase   Phrase @relation(fields: [phraseId], references: [id], onDelete: Cascade)
   /// The ID of the phrase the word is in.
   phraseId Int
-  /// The word being connecte to a phrase.
+  /// The word being connected to a phrase.
   word     Word   @relation(fields: [wordId], references: [id])
   /// The ID of the word being connecte to a phrase.
   wordId   String
--- a/packages/db/src/schema.prisma
+++ b/packages/db/src/schema.prisma
@@ -177,6 +177,8 @@ model Word {
   translatorNotes TranslatorNote[]
   // The list of footnotes in different languages for this word.
   footnotes       Footnote[]
+  /// The list of phrases this word is connected to in different languages.
+  phrases         PhraseWord[]
 
   @@index([formId])
 }
@@ -240,6 +242,8 @@ model Language {
   roles               LanguageMemberRole[]
   /// The pending import job if any.
   importJob           LanguageImportJob?
+  /// The phrases in this language.
+  phrases             Phrase[]
 }
 
 /// Keeps track of a language import job.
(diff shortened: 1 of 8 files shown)


packages/db/src/schema.prisma
```diff
@@ -301,6 +337,10 @@ model Gloss {
   gloss      String?
   /// The approval state of the gloss.
   state      GlossState @default(UNAPPROVED)
+  /// The phrase the gloss applies to.
+  phrase     Phrase?    @relation(fields: [phraseId], references: [id])
```
Pertempto, line 341: So glosses still apply to individual words, but they can also apply to an optional phrase?

packages/db/src/schema.prisma
```diff
@@ -282,6 +286,38 @@ model LanguageMemberRole {
...
+  footnote       Footnote?
+}
+
+// Connects one or more words to a phrase in a language.
+model PhraseWord {
+  /// The phrase the word is in.
+  phrase   Phrase @relation(fields: [phraseId], references: [id], onDelete: Cascade)
+  /// The ID of the phrase the word is in.
+  phraseId Int
+  /// The word being connecte to a phrase.
```
Pertempto: typo

packages/db/src/schema.prisma
arrocke: This is part of the expand/contract pattern that I laid out in the [ticket](https://github.com/globalbibletools/gbt/issues/376#issuecomment-2028517038). This PR is the first three steps, and the next PR will cover step 5, which we will release after we have completed steps 3 and 4 in production.

packages/db/src/schema.prisma
Pertempto:
```suggestion
  /// The word being connected to a phrase.
```

packages/db/src/sql/migrate-language-phrases.sql
Pertempto, line 1: When does this file get used? Should I test it?

Pertempto (approved):
I can update glosses, translator notes, and footnotes with no issues.

Pertempto:
Does `GlossHistoryEntry` need a `phraseId` field?

packages/db/src/sql/migrate-language-phrases.sql
arrocke: @Pertempto This file can be used to construct phrases for glosses and notes for a single language that don't have one. You could run it locally on a language to confirm that it sets up phrases if you'd like. Up to you though. I plan to run this on the test language in production as an initial test, and then on everything else

arrocke:
> Does GlossHistoryEntry need a phraseId field?

This will be addresed in #384

packages/db/src/sql/migrate-language-phrases.sql
Pertempto: Don't wait on me.

PR #389, opened 2024-04-24
feat: notes indication

### LINKED ISSUE #383: Indicate which words have notes in translation line
* To the left of the hebrew and right of the greek in the translation line, add the [note-sticky](https://fontawesome.com/icons/note-sticky?f=classic&s=solid) icon if the word has either a footnote or a translator note
* If the user does not have access to translator notes, only show the icon if there is a footnote
* Clicking the icon should open the note sidebar and put focus on the first text editor if the user has edit permissions

tycebrown:
## What has changed
- Added sticky note button next to words with notes 
- When clicking this sticky note button, the note tab is opened with focus on the translator editor.

## Connected Issues

closes #383 

## QA Steps

1. Select a word and add notes. Confirm that when the notes are saved, the note icon appears.
2. Close the sidebar
3. Press the sticky note. Confirm that the notes tab opens and the editor is focused.

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -118,7 +118,7 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       <li ref={root} dir={originalLanguage === 'hebrew' ? 'rtl' : 'ltr'}>
         <div
           id={`word-${word.id}`}
-          className={`flex items-center h-8 cursor-pointer font-mixed ${
+          className={`flex items-center gap-1.5 h-8 cursor-pointer font-mixed ${
             originalLanguage === 'hebrew' ? 'text-right pr-3' : 'text-left pl-3'
           }`}
           tabIndex={-1}
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -88,6 +88,14 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       () => apiClient.verses.findNotes(verseId, targetLanguage?.code ?? '')
     );
 
+    const hasTranslatorNote = !isRichTextEmpty(
+      notesQuery.data?.data?.translatorNotes[word.id].content ?? ''
+    );
+    const hasFootnote = !isRichTextEmpty(
+      notesQuery.data?.data?.footnotes[word.id].content ?? ''
+    );
+    const hasNote = hasFootnote || (hasTranslatorNote && isTranslator);
+
     const glossWidth = useTextWidth({
       text: glossValue ?? '',
       fontFamily: expandFontFamily(targetLanguage?.font ?? 'Noto Sans'),
@@ -99,21 +107,14 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
     useLayoutEffect(() => {
       setWidth(
         Math.max(
-          ancientWord.current?.clientWidth ?? 0,
+          (hasNote ? 28 : 0) + (ancientWord.current?.clientWidth ?? 0),
           refGloss.current?.clientWidth ?? 0,
           // The extra 24 pixels accommodates the google icon
           // The extra 48 pixels accommodates the approval button
           glossWidth + (hasMachineSuggestion ? 24 : 0) + 44
         )
       );
-    }, [glossWidth, hasMachineSuggestion]);
-
-    const hasTranslatorNote = !isRichTextEmpty(
-      notesQuery.data?.data?.translatorNotes[word.id].content ?? ''
-    );
-    const hasFootnote = !isRichTextEmpty(
-      notesQuery.data?.data?.footnotes[word.id].content ?? ''
-    );
+    }, [hasNote, glossWidth, hasMachineSuggestion]);
 
     return (
       <li ref={root} dir={originalLanguage === 'hebrew' ? 'rtl' : 'ltr'}>
@@ -132,11 +133,7 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
             {word.text}
           </span>
           <Button
-            className={
-              hasFootnote || (hasTranslatorNote && isTranslator)
-                ? 'inline-block'
-                : 'hidden'
-            }
+            className={`pe-0 ${hasNote ? 'inline-block' : 'hidden'}`}
             small
             variant="tertiary"
             onClick={onOpenNotes}
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -118,7 +118,7 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       <li ref={root} dir={originalLanguage === 'hebrew' ? 'rtl' : 'ltr'}>
         <div
           id={`word-${word.id}`}
-          className={`h-8 cursor-pointer font-mixed ${
+          className={`flex items-center h-8 cursor-pointer font-mixed ${
             originalLanguage === 'hebrew' ? 'text-right pr-3' : 'text-left pl-3'
           }`}
           tabIndex={-1}
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -132,7 +132,7 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
           </span>
           <Button
             className={
-              hasFootnote || (hasTranslatorNote && hasLanguageReadPermissions)
+              hasTranslatorNote || (hasFootnote && hasLanguageReadPermissions)
                 ? 'inline-block'
                 : 'hidden'
             }
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -17,6 +17,7 @@ import Button from '../../shared/components/actions/Button';
 import { useAccessControl } from '../../shared/accessControl';
 import apiClient from '../../shared/apiClient';
 import { useQuery } from '@tanstack/react-query';
+import { isRichTextEmpty } from '../../shared/components/form/RichTextInput';
 
 export interface TranslateWordProps {
   editable?: boolean;
@@ -107,9 +108,12 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       );
     }, [glossWidth, hasMachineSuggestion]);
 
-    const hasTranslatorNote =
-      !!notesQuery.data?.data?.translatorNotes[word.id].content;
-    const hasFootnote = !!notesQuery.data?.data?.footnotes[word.id].content;
+    const hasTranslatorNote = !isRichTextEmpty(
+      notesQuery.data?.data?.translatorNotes[word.id].content ?? ''
+    );
+    const hasFootnote = !isRichTextEmpty(
+      notesQuery.data?.data?.footnotes[word.id].content ?? ''
+    );
 
     return (
       <li ref={root} dir={originalLanguage === 'hebrew' ? 'rtl' : 'ltr'}>
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -132,8 +132,8 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
           </span>
           <Button
             className={
-              hasTranslatorNote || (hasFootnote && hasLanguageReadPermissions)
-                ? ''
+              hasFootnote || (hasTranslatorNote && hasLanguageReadPermissions)
+                ? 'inline-block'
                 : 'hidden'
             }
             small
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -14,6 +14,10 @@ import { capitalize } from '../../shared/utils';
 import AutocompleteInput from '../../shared/components/form/AutocompleteInput';
 import { TextDirection } from '@translation/api-types';
 import Button from '../../shared/components/actions/Button';
+import { useAccessControl } from '../../shared/accessControl';
+import { useParams } from 'react-router-dom';
+import apiClient from '../../shared/apiClient';
+import { useQuery } from '@tanstack/react-query';
 
 export interface TranslateWordProps {
   editable?: boolean;
@@ -28,6 +32,7 @@ export interface TranslateWordProps {
   onChange(data: { gloss?: string; approved?: boolean }): void;
   onFocus?: () => void;
   onShowDetail?: () => void;
+  onOpenNotes?: () => void;
 }
 
 export interface TranslateWordRef {
@@ -49,10 +54,16 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       onChange,
       onFocus,
       onShowDetail,
+      onOpenNotes,
     }: TranslateWordProps,
     ref
   ) => {
     const { t, i18n } = useTranslation(['translate']);
+    const params = useParams() as {
+      language: string;
+      verseId: string;
+    };
+    const userCan = useAccessControl();
     const input = useRef<HTMLInputElement>(null);
 
     const root = useRef<HTMLLIElement>(null);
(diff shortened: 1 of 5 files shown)


Pertempto:
The note button feels a little high compared to the text. One way to align them better would to be to add the tailwind classes `flex items-center` to the parent `div`

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto, line 1: What happened that caused such a huge diff in this file? It looks like a bunch of lines got changed that didn't need it?

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -29,295 +36,315 @@ type TranslationSidebarProps = {
...
-  };
+    const [tabIndex, setTabIndex] = useState(0);
 
-  return (
-    <div
-      className={`
+    useImperativeHandle(ref, () => ({
+      openNotes: () => {
+        flushSync(() => setTabIndex(1));
+        translatorNotesRef.current?.focus();
```
Pertempto: The focus isn't working unless I press the button again after the tab has been changed. What is the `flushSync` doing? I think the problem is that we are calling `translatorNotesRef.current?.focus();` too quickly. I'm guessing that if you put a debug message in there, you'd find that `translatorNotesRef.current` is `null` when you click the open note button the first time with the sidebar closed.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -107,6 +130,18 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
           <span className="inline-block" ref={ancientWord}>
             {word.text}
           </span>
+          <Button
+            className={
+              hasFootnote || (hasTranslatorNote && hasLanguageReadPermissions)
```
Pertempto: I think this logic might be a little mixed up. If I'm understanding the issue description correctly, it should be `hasTranslatorNote || (hasFootnote && hasLanguageWritePermission)`.

@arrocke maybe I'm misreading it, please correct me if I'm wrong.

tycebrown: @Pertempto I think I did mess up the order there. Sorry about that.

tycebrown: Done.

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Oops, I forgot to push my changes. Sorry about that. Done.

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Since I put the `TranslationSidebar` in a `forwardRef`, it increased the indentation of the body of the function. I pushed a commit to make the diff look a little nicer, but it still has a lot of changes. You could try viewing the diff in vscode. Should I try to edit the formatting further?

Pertempto: Ahh the `forwardRef` did it. No, don't worry about adjusting formatting. It'll be fine. Maybe you could point me to the sections of code in this file where you did the actual changes, so I can focus on those?

packages/web/src/features/translation/TranslateWord.tsx
Pertempto: We also want to check the write permission, not the read permission, correct?

Pertempto: Now that I re-read Adrian's issue description, I think maybe your logic was correct, it just needed to check write permissions instead of read permissions. I'm getting myself confused now 😆 

I think it would make things more clear if you renamed `hasLanguageReadPermissions` to `isTranslator` (and check the write permission, not the read permission)

After that, I think we'd write the logic like this:


`hasFootnote || (hasTranslatorNote && isTranslator)`

Anyone can view the foot notes, only translators can view the translator note

Pertempto: @arrocke please correct me if I'm wrong again.

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Points I made changes:
- Before the return statement, I created the `tabIndex`  and `shouldFocusNotes` state variables. 
  - I also added a `useImperativeHandle` hook which exposes the `openNotes` function.
 - The translator notes `RichTextInput` was modified
   - The `autoFocus` prop was added
   - In the `onBlur` callback, a call to `setShouldFocusNotes` was added.

Pertempto: Looks good!

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: It works great now! 👍🏽

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -91,11 +110,15 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
...
+    const hasTranslatorNote =
+      !!notesQuery.data?.data?.translatorNotes[word.id].content;
+    const hasFootnote = !!notesQuery.data?.data?.footnotes[word.id].content;
+
     return (
       <li ref={root} dir={originalLanguage === 'hebrew' ? 'rtl' : 'ltr'}>
         <div
           id={`word-${word.id}`}
-          className={`h-8 cursor-pointer font-mixed ${
+          className={`flex items-center h-8 cursor-pointer font-mixed ${
```
Pertempto:
```suggestion
          className={`flex items-center gap-1 h-8 cursor-pointer font-mixed ${
```

Maybe we should add `gap-2` or `gap-1` here, so that the button isn't immediately beside the text. Fiddle around with it to figure out which size of gap looks best.

image

Pertempto (approved):
It functions great! Nice work!

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: @Pertempto I've settled right in the middle with `gap-1.5` :) ~~I'll merge once I can commit the changed~~

packages/web/src/features/translation/TranslateWord.tsx
arrocke: Yes this is the right idea. We don't want to expose translator notes to users who can't view them. They are meant for internal communication while footnotes are for the general user. Using the read permission is probably better because a language admin can view everything even if they don't have permissions to write notes

arrocke:
I'll try to get a review on this within the next few days.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -49,10 +54,20 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
...
       onShowDetail,
+      onOpenNotes,
     }: TranslateWordProps,
     ref
   ) => {
     const { t, i18n } = useTranslation(['translate']);
+    const params = useParams() as {
+      language: string;
+      verseId: string;
+    };
```
arrocke: We are already passing the language code in `targetLanguage`, so we can add that to the props interface and then use that instead of params. That's preferred because it doesn't couple this component to the URL

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -91,11 +110,15 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       );
     }, [glossWidth, hasMachineSuggestion]);
 
+    const hasTranslatorNote =
+      !!notesQuery.data?.data?.translatorNotes[word.id].content;
+    const hasFootnote = !!notesQuery.data?.data?.footnotes[word.id].content;
```
arrocke: We need to use tiptap to check if content is empty because if you delete the content of a note, there will still be HTML inside content and this will be true. The tiptap editor has an [isEmpty](https://tiptap.dev/docs/editor/api/editor#is-empty) getter that we can use for this purpose. Maybe we can export an `isRichTextEmpty` function from the RichTextInput file that takes an HTML string and determines if it is empty.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -91,11 +110,15 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       );
     }, [glossWidth, hasMachineSuggestion]);
```
arrocke: We need to modify this effect that computes the width of the of the word to account for the note icon. The first argument to Math.max needs to include the width of the note icon and spacing if the note is present. This allows the entire word component to expand in width if that top line is longest.

packages/web/src/shared/components/form/RichTextInput.tsx
```diff
@@ -71,6 +75,10 @@ const RichTextInput = forwardRef<RichTextInputRef, RichTextInputProps>(
       });
     }, [value, editor]);
 
+    useEffect(() => {
+      if (autoFocus) editor?.commands.focus();
+    }, [autoFocus, editor]);
```
arrocke: We should use `useImperativeHandle` to implement the `focus` function of the ref. Then you don't need all of the hoops to keep track of the focus state in the sidebar. Just grab a ref to the note input and then execute focus in the imperative handle hook

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -464,6 +476,7 @@ export default function TranslationView() {
                         }}
                         onFocus={() => setSidebarWordIndex(i)}
                         onShowDetail={() => setShowSidebar(true)}
+                        onOpenNotes={() => setShouldOpenNotes(true)}
```
arrocke: Here you can just execute `sidebarRef.current.openNotes()` directly. No need to create additional state and effects.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -12,312 +20,328 @@ import LoadingSpinner from '../../shared/components/LoadingSpinner';
...
-    previewElement.style.backgroundColor = '#ffffff80';
-    previewElement.style.float = 'left';
-    anchorElement.insertAdjacentElement('afterend', previewElement);
-    setPreviewElement(previewElement);
-  };
+    const [shouldFocusNotes, setShouldFocusNotes] = useState(false);
+    useImperativeHandle(ref, () => ({
+      openNotes: () => {
+        setTabIndex(1);
+        setShouldFocusNotes(true);
```
arrocke: If here we are calling focus on the rich text input ref, we may need to use setTimeout with no delay to force the focus call to occur after the browser has had a chance to repaint after we've switched the active tab

arrocke (changes requested):

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: Done. I also moved the verseId URL param to the props.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -70,6 +82,10 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
       glossValue ?? ''
     );
     const hasMachineSuggestion = !gloss && !suggestions[0] && !!machineGloss;
+    const notesQuery = useQuery(
+      ['verse-translator-notes', targetLanguage?.code ?? '', verseId],
+      () => apiClient.verses.findNotes(verseId, targetLanguage?.code ?? '')
+    );
```
arrocke: I'm just realizing that perhaps we should lift this to the view and pass a `hasNote` prop so that we only try to load it once and not for each word

tycebrown:
@arrocke I'll be able to get to this on Saturday.

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: Done. ~~There was some useless padding at the end of the note button so I added `pe-0`~~

packages/web/src/features/translation/TranslationView.tsx
tycebrown: Done. I ended up wrapping this call in a `setTimeout` as well for the case where the user does not have a sidebar open (so `sidebarRef.current` is `null`)

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: Done.

packages/web/src/shared/components/form/RichTextInput.tsx
tycebrown: I'm having trouble getting this to work consistently. Could you review my code to see what I am doing wrong?

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -107,6 +112,14 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
           <span className="inline-block" ref={ancientWord}>
```
arrocke: Can we move the tabIndex and click event to this span instead? Then we can call `onShowDetail` from the button click as well as `onOpenNotes`. This is a bit more explicit as to what is different about each click region.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -107,6 +112,14 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
...
             {word.text}
           </span>
+          <Button
+            className={hasNote ? 'inline-block' : 'hidden'}
+            small
+            variant="tertiary"
+            onClick={onOpenNotes}
```
arrocke: We are going to want a `tabIndex` of -1 so that this isn't in the tab sequence. That will be tedious to navigate through each note icon.

packages/web/src/shared/components/form/RichTextInput.tsx
arrocke: Looks like the issue is that there is a race condition between the fall to focus, and the creation of the editor state. This means that when focus is called, if the rich text editor has not mounted yet, the editor will be null for long enough that focus doesn't do anything. We can fix this by adding `unmount={false}` to each `Tab.Panel` in the sidebar so that the tabs are always mounted.

arrocke (changes requested):
Nice work so far! This is getting close

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: Done.

tycebrown:
@arrocke After spending way too much time on this, I found that adding `unmount={false}` to the `Tab.Panel`s partially fixed the issue, but, switching focus between different words inconsistnelty failed. I believe the problem is due to the fact that the `key` of the editor changes between renders (see [here](https://react.dev/learn/rendering-lists#why-does-react-need-keys)). Also, using console log statements I found that the editor was not null (at least with `unmount={false}`, but the problem was still there. 

So, I ended up moving the focus effect to the rich text input. 

I am sorry I have dragged this trivial ticket out to take a week's time. I've been thinking about following Mr. Bruno's suit and sharpening my web development/programming skills.

Pertempto:
> sharpening my web development/programming skills.

I think that you'll find that practicing on a real project like this **is** the best way to sharpen your skills.

packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -82,31 +86,48 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
...
+          >
             {word.text}
           </span>
+          <Button
+            className={hasNote ? 'inline-block' : 'hidden'}
+            small
+            variant="tertiary"
+            tabIndex={-1}
+            onClick={() => {
+              onFocus?.();
```
arrocke, line 124: We don't need to call onFocus when clicking the notes button because focus is going to move to the notes input

arrocke (changes requested):
> I believe the problem is due to the fact that the key of the editor changes between renders (see [here](https://react.dev/learn/rendering-lists#why-does-react-need-keys)).

This is a good catch! The key change forces react to construct a new component from scratch, which then requires that tiptap generate a new editor instance after the component mounts. We actually don't need a key here because the RichTextInput component is designed to replace it's content when the `value` prop changes. So I think we can remove the keys, use `unmount={false}` and call focus directly in the imperative handle hook again.

> I am sorry I have dragged this trivial ticket out to take a week's time. I've been thinking about following Mr. Bruno's suit and sharpening my web development/programming skills.

Addison is right. Struggle is necessary for learning. This ticket isn't super urgent, so taking time to struggle through is not a problem. It's my job to manage the timeline, so I'll communicate when there is urgency. Keep up the great work!

packages/web/src/features/translation/TranslateWord.tsx
tycebrown: @arrocke I'm a little confused here. Doesn't `onFocus` set the word id? Conceptually, would that still make sense?

arrocke: Oh good point, I missed that.

arrocke (approved):

PR #393, opened 2024-04-29
fix: blurring notes editor leads to loss of data

### LINKED ISSUE #392: Bug: blurring the notes editor can lead to loss of data
**Describe the bug**
When the user switches focus from the notes editor of one word to another and then blurs, the notes of the second word will contain the contents of the first.

**To Reproduce**

1. Find two words with translator notes.
   - This bug can be reproduced with footnotes as well 
2. Open the notes tab for the first word. Focus on translator notes editor and observe its content.
3. Open the notes tab for second word. Focus on translator notes editor and observe its content.
4. Blur. Reload the page. 
5. Confirm that the notes for the first word have been copied to (or swapped with) the second.
  

**Expected behavior**
Notes for the second word should not have changed, neither should the notes for the first word.

**Screenshots**
Here is a demonstration of the bug:

https://github.com/globalbibletools/gbt/assets/91034018/054deaf6-f9dc-4142-abca-0964ee253d09

- Device: Laptop
- OS: Window
- Browser: brave (chrome-based)
- Version: v1.65

**Additional context**
I think the problem lies in the hidden input in `RichTextInput`.

tycebrown:
## What has changed
- The bug was caused by the fact that the `value` in the hidden input of the `RichTextInput` was not updated when the `value` prop changed. So, when the user switched words, the hidden input of the notes editor was not updated and `saveTranslatorNote` would save the notes of the previous word to the database.
-  Now, changing the `value` prop of the `RichTextInput` emits an update. This updates the value in the hidden input. 

## Connected Issues

closes #392

## QA Steps

1. Find two words with translator notes (or footnotes)
2. Open the notes tab for the first word. Focus on translator notes editor and observe its content.
3. Open the notes tab for second word. Focus on translator notes editor and observe its content.
4. Blur. Reload the page.
5. Confirm that the notes of the first and second words have not changed.

## Post-Deployment

- [ ] Nothing required

### DIFF
--- a/packages/web/src/shared/components/form/RichTextInput.tsx
+++ b/packages/web/src/shared/components/form/RichTextInput.tsx
@@ -1,16 +1,15 @@
 import { useEditor, EditorContent } from '@tiptap/react';
 import StarterKit from '@tiptap/starter-kit';
 import { Icon } from '../Icon';
-import { ComponentProps, forwardRef, useEffect, useRef } from 'react';
+import { ComponentProps, forwardRef, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
-import { ChangeHandler } from 'react-hook-form';
 
 export interface RichTextInputProps {
   name: string;
   value?: string;
   defaultValue?: string;
-  onChange?: ChangeHandler;
-  onBlur?: ChangeHandler;
+  onChange?: (newValue: string) => void;
+  onBlur?: () => void;
   'aria-labelledby'?: string;
   'aria-label'?: string;
 }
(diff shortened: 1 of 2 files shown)


Pertempto (approved):
I first tested on `main` so that I could verify I had the correct test steps. Then I tested this branch. It works great! Nice work!

packages/web/src/shared/components/form/RichTextInput.tsx
```diff
@@ -66,7 +66,7 @@ const RichTextInput = forwardRef<RichTextInputRef, RichTextInputProps>(
     });
 
     useEffect(() => {
-      editor?.commands.setContent(value ?? '', false, {
+      editor?.commands.setContent(value ?? '', true, {
```
Pertempto: Did you do additional testing to make sure this change doesn't break anything else? I wonder why we used `false` in the first place.

arrocke: We don't want to emit events when you update the internal state of the rich text editor when props change, only when the state changed through user input. That's why this is set to false.

I don't think we need the hidden input anymore, so I'd just modify the change event to return a string and the blur event to not have any arguments. That will just simplify things overall.

arrocke (approved):

PR #400, opened 2024-05-12
feat: saving indicator in notes tab

### LINKED ISSUE #395: Saving indicator for notes
Both kinds of notes when they are being saved to the database should show a saving icon and the text `Saving` next to the header

tycebrown:
## What has changed
New saving indicator which appears when a note is being saved.
## Connected Issues
closes #395 

## QA Steps
1. Open the sidebar for a word. 
2. Edit the translator notes and blur. Confirm that the saving indicator appears.
3. Repeat for footnotes.

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslationSidebar.tsx
+++ b/packages/web/src/features/translation/TranslationSidebar.tsx
@@ -285,9 +285,9 @@ export const TranslationSidebar = forwardRef<
                           {t('translate:translator_notes')}
                         </h2>
                         {isSavingTranslatorNote && (
-                          <em>
+                          <span className="italic">
                             <Icon icon="save" /> {t('common:saving')}...
-                          </em>
+                          </span>
                         )}
                       </div>
                       {translatorNote?.authorName && (
@@ -319,9 +319,9 @@ export const TranslationSidebar = forwardRef<
                     <div className="flex flex-row gap-2.5">
                       <h2 className="font-bold">{t('translate:footnotes')}</h2>
                       {isSavingFootnote && (
-                        <em>
+                        <span className="italic">
                           <Icon icon="save" /> {t('common:saving')}...
-                        </em>
+                        </span>
                       )}
                     </div>
                     {hasLanguageReadPermissions && footnote?.authorName && (
--- a/packages/web/src/features/translation/TranslationSidebar.tsx
+++ b/packages/web/src/features/translation/TranslationSidebar.tsx
@@ -290,7 +290,7 @@ export const TranslationSidebar = forwardRef<
                         </h2>
                         {saveTranslatorNoteMutation.isLoading && (
                           <em>
-                            <Icon icon="save" /> Saving...
+                            <Icon icon="save" /> {t('common:saving')}...
                           </em>
                         )}
                       </div>
@@ -324,7 +324,7 @@ export const TranslationSidebar = forwardRef<
                       <h2 className="font-bold">{t('translate:footnotes')}</h2>
                       {saveFootnoteMutation.isLoading && (
                         <em>
-                          <Icon icon="save" /> Saving...
+                          <Icon icon="save" /> {t('common:saving')}...
                         </em>
                       )}
                     </div>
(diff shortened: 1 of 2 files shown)


packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -103,39 +103,48 @@ export const TranslationSidebar = forwardRef<
...
-              note: noteContent,
-            });
+            await saveTranslatorNoteMutation.mutateAsync(noteContent);
             notesQuery.refetch();
           },
           15000,
           { leading: false, trailing: true }
         ),
-      [language, notesQuery, word.id]
+      [notesQuery, saveTranslatorNoteMutation]
```
Pertempto: @arrocke is it safe to make this change to the dependencies array?

Pertempto (approved):
It works great!

packages/web/src/features/translation/TranslationSidebar.tsx
arrocke: `saveTranslatorNoteMutation` is going to change whenever the state of the mutation changes, we want to only rebuild this function when `mutateAsync` changes so we should destructure it before using it. Something like:
```ts
const {
  isLoading: isSavingTranslatorNote,
  mutateAsync: mutateTranslatorNote
} = useMutation()
```
We can also move `notesQuery.refetch()` into `onSuccess` of `useMutation` and then we don't need that as a dependency either

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -275,9 +284,16 @@ export const TranslationSidebar = forwardRef<
...
                     <div className="flex flex-col gap-2">
-                      <h2 className="font-bold">
-                        {t('translate:translator_notes')}
-                      </h2>
+                      <div className="flex flex-row gap-2.5">
+                        <h2 className="font-bold">
+                          {t('translate:translator_notes')}
+                        </h2>
+                        {saveTranslatorNoteMutation.isLoading && (
+                          <em>
```
arrocke: It's better to use `<span class="italic">` than an `em` because `em` has semantic meaning that doesn't apply here

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -103,39 +103,48 @@ export const TranslationSidebar = forwardRef<
         );
       }
     }, [word.id, notesQuery]);
+    const saveTranslatorNoteMutation = useMutation({
+      mutationFn: async (noteContent: string) =>
+        await apiClient.words.updateTranslatorNote({
+          wordId: word.id,
+          language,
+          note: noteContent,
+        }),
```
arrocke: I don't think the mutation function should capture variables from outside it's scope. We should make the argument to the mutation be an object of `wordId`, `language`, and `note`, and then pass those in when executing

arrocke (changes requested):

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -103,39 +103,44 @@ export const TranslationSidebar = forwardRef<
...
+      [language, mutateTranslatorNote, word.id]
     );
 
+    const { isLoading: isSavingFootnote, mutateAsync: mutateFootnote } =
+      useMutation({
+        mutationFn: (data: {
+          wordId: string;
+          language: string;
+          note: string;
+        }) => apiClient.words.updateFootnote(data),
```
tycebrown, line 132: @arrocke Would this be a good place to use `useCallback`? What about for `onSuccess`?

arrocke (approved):

packages/web/src/features/translation/TranslationSidebar.tsx
arrocke: I don't think it's necessary here

PR #402, opened 2024-05-14
fix: notes editor update issue

### LINKED ISSUE #401: bug: notes editor update issue
**Describe the bug**
Let's say there are two words which don't have any notes. If the user adds notes to the first word, and then switches focus to the second word, the notes editor will still display the notes added to the first word.

**To Reproduce**

1. Find two words without notes.
2. Open the notes tab for the first word. Enter any content into one of the notes editors.
3. Switch focus to the second word.
4. Confirm that the editor still has the notes content from the first word.

**Expected behavior**
The editor of the second word should be empty. 

- Device: Laptop
- OS: Windows
- Browser: Brave 
- Version: 1.65.132

**Screenshots**

https://github.com/globalbibletools/gbt/assets/91034018/caf0abf3-582d-4494-a768-894dd6fa2f86

tycebrown:
## What has changed
- Fixed the bug described in #401
   - The method of updating the editor's content has switched from using state to using refs.

## Connected Issues
closes #401 

## QA Steps
1. Find two words without notes.
2. Open the notes tab for the first word. Enter any content into one of the notes editors.
3. Switch focus to the second word.
4. Confirm that the editor is empty.
5. Now enter content into the editor of the second word.
6. Switch the translation language to English. Confirm that the editor is empty (or contains only the english notes)

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslationSidebar.tsx
+++ b/packages/web/src/features/translation/TranslationSidebar.tsx
@@ -89,20 +89,40 @@ export const TranslationSidebar = forwardRef<
       id: language,
     });
 
-    const [translatorNoteContent, setTranslatorNoteContent] = useState('');
+    // const [translatorNoteContent, setTranslatorNoteContent] = useState('');
     const [footnoteContent, setFootnoteContent] = useState('');
     const wordId = useRef('');
+
+    const [isNotesTabMounted, setIsNotesTabMounted] = useState<boolean>(false);
+    const translatorNotesEditorRef = useRef<RichTextInputRef>(null);
+
     useEffect(() => {
-      if (notesQuery.isSuccess && word.id !== wordId.current) {
+      console.log(`################### mounted? ${isNotesTabMounted}`);
+      if (
+        isNotesTabMounted &&
+        notesQuery.isSuccess &&
+        word.id !== wordId.current
+      ) {
         wordId.current = word.id;
-        setTranslatorNoteContent(
-          notesQuery.data.data.translatorNotes[word.id]?.content ?? ''
+        console.log(
+          `&&&&&&&&&&&&&&&&&&&&& notes eidtor ref: ${translatorNotesEditorRef.current}`
+        );
+        console.log(
+          `********************* notes query content: ${
+            notesQuery.data.data.translatorNotes[word.id]?.content
+          }`
+        );
+        translatorNotesEditorRef.current &&
+          (translatorNotesEditorRef.current.value =
+            notesQuery.data.data.translatorNotes[word.id]?.content ?? '');
+        console.log(
+          `^^^^^^^^^^^^^^^^^^^^^ notes editor value now: ${translatorNotesEditorRef.current?.value}`
         );
         setFootnoteContent(
           notesQuery.data.data.footnotes[word.id]?.content ?? ''
         );
       }
-    }, [word.id, notesQuery]);
+    }, [word.id, notesQuery, isNotesTabMounted]);
 
     const saveTranslatorNote = useMemo(
       () =>
@@ -148,6 +168,15 @@ export const TranslationSidebar = forwardRef<
     );
     const [previewVerseIds, setPreviewVerseIds] = useState<string[]>([]);
 
+    useImperativeHandle(ref, () => ({
+      openNotes: () => {
+        setTabIndex(1);
+        setTimeout(() => {
+          translatorNotesEditorRef.current?.focus();
+        }, 0);
+      },
+    }));
+
     useEffect(() => {
       const { current } = lexiconEntryRef;
       // Highlight references to the currently selected verse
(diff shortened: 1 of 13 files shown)


packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -294,12 +307,16 @@ export const TranslationSidebar = forwardRef<
                         <RichTextInput
                           ref={translatorNotesEditorRef}
                           name="translatorNoteContent"
-                          value={translatorNoteContent}
                           onBlur={() => saveTranslatorNote.flush()}
                           onChange={saveTranslatorNote}
```
arrocke: The key problem is that when the rich text content changes, we aren't updating `translatorNoteContent` and `footnoteContent`. Here is why that is a problem:

1. Find a verse with no notes. In this case whenever you switch words, `footnoteContent` is set to the empty string.
2. Type a footnote. The rich text input's state as well as the server state have the content of the note, but `footnoteContent` does not.
3. Switch words to one without a note. Now the sidebar will reset `footnoteContent` to the contents of the note for that word which is the same empty string. If it had changed, then the rich text input would reset with the new content, but in this case it didn't because the value being passed didn't change

If we change this onChange handler to something like the following, the issue is resolved:
```ts
onChange={content => {
  setTranslatorNoteContent(content)
  saveTranslatorNote(content)
})
```
Now whenever you type in the rich text input, the content states will change in the sidebar so that when they are reset when the word changes, the rich text input actually notices that the content changes and it needs to reset

arrocke (changes requested):
Using refs like we are here is a code smell. We shouldn't need imperative code to handle state in a react app. Below is an explanation of a better way to solve this.

packages/web/src/features/translation/TranslationSidebar.tsx
Pertempto: Thanks for the detailed explanation!

tycebrown: ~~Done.~~

tycebrown: Alright, done now.

arrocke (commented):
There are a lot of changed lines in here that now don't need to change, can we eliminate those? That makes the git history of these files easier to parse

tycebrown:
@arrocke is that better?

packages/web/src/shared/components/RichText.tsx
```diff
@@ -22,7 +22,7 @@ export default function RichText({ content, className = '' }: RichTextProps) {
     editor?.commands.setContent(content, false, {
       preserveWhitespace: 'full',
     });
-  }, [editor?.commands, content]);
```
arrocke: This line needs to stay the same. It ensures this effect is only executed if the commands object changes

packages/web/src/features/translation/TranslationSidebar.tsx
```diff
@@ -91,18 +91,18 @@ export const TranslationSidebar = forwardRef<
 
     const [translatorNoteContent, setTranslatorNoteContent] = useState('');
     const [footnoteContent, setFootnoteContent] = useState('');
-    const wordId = useRef('');
+
     useEffect(() => {
-      if (notesQuery.isSuccess && word.id !== wordId.current) {
-        wordId.current = word.id;
```
arrocke: We still need these changes. It ensures we only reset the note state when the word id changes

arrocke (changes requested):

packages/web/src/shared/components/RichText.tsx
tycebrown: Done.

packages/web/src/features/translation/TranslationSidebar.tsx
tycebrown: @arrocke I'm a little confused. As I understand, `notesData` (which is just `notesQuery.data`) will only change when:
1. The `language` changes
2. The `word.id` changes
3. New data is fetched

Wouldn't we want to update the notes content in all these cases (or at least 1 and 2)? It seems like the only reason we needed the `wordId` ref was because the `notesQuery` object, which was used as a dependency instead of the `notesData`, changes every render.

arrocke: We are trying to prevent the 3rd case. After autosave, we fetch the latest version of the notes, but that might be a few characters behind the current value of the text input because you can keep typing while the network requests settle. The wordId ref prevents the notes from being reset if the update was due to an autosave rather than a word change. This works for word changes, but not for languages. That's ok for now because I accidentally fixed this by introducing phrases to the frontend on a separate branch. So for now, let's leave this as it was

tycebrown: Done @arrocke

arrocke (approved):

Issue #361, opened 2024-02-27
Combine multiple ancient words into a single gloss


arrocke:
For [agglutinative languages](https://en.wikipedia.org/wiki/Agglutinative_language), we need to be able to combine multiple Hebrew or Greek words together into a phrasal gloss. This grouping, will share a gloss, notes, and comments.

arrocke:
Idea here is that a shift click/enter allows you to select multiple words. Then the link words button connects the words into a single phrase. At this point the state of the gloss for one word is the same as the others. When focused on a linked word, the link words button changes to unlink. We could also show a link icon next to the word when it is linked to another.

![Image](https://github.com/globalbibletools/gbt/assets/11968925/9053fb5c-2223-4ffc-9a77-f065ff27ab9a)

arrocke:
Next step is to figure out the API routes we need to build to support this.

One idea:
`GET /languages/{code}/verses/{verseId}/[glosses|notes]` returns a list of phrases rather than words
`POST /languages/{code}/phrases` creates a new phrase from a list of words
`DELETE /languages/{code}/phrases/{phraseId}` unlinks a phrase, removing associated gloss and notes
`PUT /languages/{code}/phrases/{phraseId}/gloss` updates the gloss for a phrase
`PUT /languages/{code}/phrases/{phraseId}/translator-note` updates the translator note for a phrase
`PUT /languages/{code}/phrases/{phraseId}/footnote` updates the footnote for a phrase
`POST /languages/{code}/glosses/bulk` updates glosses for multiple phrases. A phraseId updates the phrase, and a list of wordIds creates a phrase with a gloss

arrocke:
Another idea is to break up the GET routes like this:
`GET /languages/{code}/verses/{verseId}/phrases` - returns the gloss and notes for each phrase in the verse
`GET /languages/{code}/verses/{verseId}/suggestions` - returns the machine gloss and list of suggestions for each word in the verse

We could break up the suggestions route further:
`GET /languages/{code}/verses/{verseId}/machine-suggestions` - returns just the google translate results
`GET /languages/{code}/lemmas/{lemmaId}/form/{formId}/suggestions` - returns the list of suggestions for the word. This could be loaded when a word is focused rather than on page load

PR #404, opened 2024-05-21
refactor: load translation data by phrase rather than by word

arrocke:
## What has changed

* Fetching notes and glosses for a verse have been replaced by fetching phrases and suggestions. These two are split because glosses and notes are associated with phrases, and suggestions are associated with individual words and languages.
* The reference gloss in english has been moved to the verse data query.
* The translation view has been updated so that data is loaded via the phrase rather than the wordId. This should allow this view to work once we can create phrases from multiple words.
* I also removed the natural language translation from the prefetch since that is not necessarily needed right away when the next verse loads

## Connected Issues

first step for #361

## QA Steps

* Confirm that translation line loads correctly
* Confirm that notes load correctly
* Confirm that glosses and notes can be updated

## Post-Deployment

- [x] Nothing required


tycebrown:
As far as I can tell, the functionality works great. I should be able to fully review this tomorrow.

packages/api-types/src/index.ts
```diff
@@ -184,17 +186,39 @@ export const GlossSource = makeEnum({
...
+  data: VersePhrase[];
+}
+
+export interface VerseWordSuggestion {
   wordId: string;
-  gloss?: string;
   suggestions: string[];
-  state: GlossState;
   machineGloss?: string;
 }
```
tycebrown, line 216: Will we eventually replace `VerseWordSuggestion` with a `VersePhraseSuggestion`? So the `suggestions` and `machineGloss` would be connected to a phrase.

packages/api/pages/api/languages/[code]/verses/[verseId]/phrases.ts
```diff
@@ -0,0 +1,118 @@
...
+      const language = await client.language.findUnique({
+        where: {
+          code: req.query.code,
+        },
+      });
+      if (!language) {
+        return res.notFound();
+      }
+
+      // TODO: hide translator notes when not a member of the language
```
Comment (tycebrown) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/phrases.ts]: Do we need to do this, since the notes are hidden when they need to be in the ui?

packages/api/pages/api/languages/[code]/verses/[verseId]/suggestions.ts
```diff
@@ -91,7 +89,7 @@ export default createRoute<{ code: string; verseId: string }>()
       const words = await client.$queryRaw<WordQuery[]>`
         SELECT
           w.id AS "wordId",
-          target.gloss AS "gloss", target.state AS "state",
+          target.gloss AS "gloss",
```
Comment (tycebrown) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/suggestions.ts:92]: It doesn't look like we use the `gloss` column or the `target` subquery anywhere. Could we remove line `target.gloss` from the query?

packages/api/pages/api/languages/[code]/verses/[verseId]/suggestions.ts
Comment (tycebrown) [inline packages/api/pages/api/languages/[code]/verses/[verseId]/suggestions.ts:1]: On lines 118 - 124:
```sql
        LEFT JOIN LATERAL (
          SELECT phw."wordId", g.gloss, g.state FROM "PhraseWord" AS phw
          JOIN "Phrase" AS ph ON ph.id = phw."phraseId"
            AND ph."languageId" = ${language.id}::uuid
          JOIN "Gloss" AS g ON g."phraseId" = ph.id
          WHERE phw."wordId" = w.id
        ) AS target ON true
```
Correct me if I'm wrong, but I think we could remove this join? I doesn't look like we really use the `target` subquery anywhere (see my other comment)

tycebrown (approved):
I did not find any errors. I have just two questions and one possible optimization, but nothing serious.

packages/api/pages/api/languages/[code]/verses/[verseId]/phrases.ts
Comment (Pertempto) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/phrases.ts]: There's no reason to send their data to the frontend if we don't need it.

Pertempto (changes requested):
Something weird is going on with the original language words.

Here is John 1:1 on production

image

Here it is locally

image

tycebrown:
> Something weird is going on with the original language words.

It looks like the api path `/verses/[verseId]` does not order the words according to their order in the verse:


image

packages/api-types/src/index.ts
arrocke: Probably not in the short term. There are a number of problems with with suggestions and phrases, that may not be worth solving at the moment:
* Suggestions are generated by grabbing all of the glosses for phrases that contain words with the same lemma form. It's not clear how to generate suggestions for phrases of more than one word that would be useful to a translator.
* Machine translations have a similar problem. When a phrase has multiple words, there isn't a clear way for us to generate an english string that we can send to google translate that would produce a useful gloss.

packages/api/pages/api/languages/[code]/verses/[verseId]/suggestions.ts
Comment (arrocke) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/suggestions.ts:92]: We are using the target gloss in `generateMachineGlossesForVerse` to determine if we need to generate a machine gloss or not. If there is already a gloss, then we don't need to bother spending money to use the google translate api to generate a machine gloss

packages/api/pages/api/languages/[code]/verses/[verseId]/phrases.ts
Comment (arrocke) [reply packages/api/pages/api/languages/[code]/verses/[verseId]/phrases.ts]: I think at this point the complexity isn't worth it. There isn't a real security concern with exposing that over the API, so I'm going to skip

arrocke:
Thanks for catching that word ordering issue. I missed the order by clause in the query. Should be fixed now

Pertempto:
Now I'm getting this error from what seems to be the `suggestions` endpoint

```
PrismaClientKnownRequestError:
Invalid `prisma.$queryRaw()` invocation:
Raw query failed. Code: `42601`. Message: `ERROR: syntax error at or near "JOIN"`
```

I don't think I'll be available the next few days to review this PR, please don't wait on me!

Issue #384, opened 2024-04-17
Migrate gloss history data structures to support phrasal glosses


arrocke:
## Description
With multi word gloss, we need to connect gloss history to phrases rather than words. This PR addresses the data migration to make this happen.

## Process
We will do this in two steps. It is simpler than #376 because nothing is reading these events yet.

1. Add new event schema, update history writes to also write to new schema, and write migration for existing events.
2. Update history writes to no longer write to the old schema, and remove the old schema from the database.

## DB Structure
```mermaid
erDiagram
  GlossEvent ||--o{ GlossEventType : ""
  User }o--o{ GlossEvent : ""
  Phrase }o--o{ GlossEvent : ""
 
  GlossEvent {
    int id PK
    int phraseId FK
    UUID userId FK
    Date timestamp
    int eventTypeId FK
    JSON data
  }

  GlossEventType {
    int id
    string name "WordAdded | WordRemoved | GlossChanged | GlossImported | PhraseDeleted"
  }
```

## Event Data Schema
```ts
interface WordAddedData {
  wordId: string
}

interface WordRemovedData {
  wordId: string
}

// undefined fields indicate no change
interface GlossChangedData {
  gloss?: string
  state?: GlossState
}

interface PhraseDeletedData {}

interface GlossImportedData {
  wordIds: string[]
  gloss: string
  state: GlossState
}
```

PR #405, opened 2024-05-25
refactor: update glosses and notes using a phrase id

arrocke:
## What has changed

* Create phrases for each word in a verse when loading the translation view
* Update the gloss and notes associated with a phrase ID rather than a word ID

## Connected Issues

next step for #361 

## QA Steps

1. Try updating a gloss, and both kinds of notes.
2. Try approval action
3. Check that gloss history entry records are created for each action

## Post-Deployment

- [x] Nothing required


tycebrown (approved):
I did not find any errors.

I did notice that focusing in and out of a word's gloss input creates an empty, unapproved gloss for that word's phrase. This is on the main branch. It did not affect functionality negatively.

Finally, one question I have: Will we, in a future PR, change the `GlossHistoryEntry` to be associated with phrases instead of words?

arrocke:
> Will we, in a future PR, change the GlossHistoryEntry to be associated with phrases instead of words?

This is defined in #384. I'm going to work on that next

PR #406, opened 2024-06-09
refactor: track changes to phrases and glosses

### LINKED ISSUE #384: Migrate gloss history data structures to support phrasal glosses

arrocke:
## What has changed

* phrases track the time of creation and deletion, as well as who created and deleted the phrase
* all queries filter out deleted queries - we keep them so we can restore them if there were deleted on accident, and so we can see who deleted
* gloss history entries were renamed to gloss events and now associate with the phrase rather than a specific word

I ended up deviating a bit from the initial design because it made things a littler simpler to implement. Because we don't need to modify the words in a phrase once it is created, we can just track the created and deleted date on the phrase itself. That negates the need for a more complicated event system for glosses since the current system will work just fine with phrase IDs rather than word and language IDs

## Connected Issues

closes #384 

## QA Steps

1. Run `nx run db:prisma migrate dev` to update your database with the new schema
2. Try updating a few glosses to see what happens as phrases and gloss events are created

## Post-Deployment

- [x] Nothing required


packages/db/src/migrations/20240609000711_phrase_gloss_events/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+ALTER TABLE "GlossHistoryEntry" RENAME TO "GlossEvent";
+
+ALTER TABLE "GlossEvent"
+  ADD COLUMN "id" SERIAL PRIMARY KEY,
+  ADD COLUMN "phraseId" INTEGER;
+
+UPDATE "GlossEvent"
+  SET "phraseId" = data."phraseId"
+FROM (
+  SELECT e."languageId", e."wordId", ph.id AS "phraseId"
```
arrocke, line 23: This should set the phrase IDs of all gloss events before making it a required field

packages/db/src/migrations/20240609010830_require_phrase_created_date/migration.sql
```diff
@@ -0,0 +1,17 @@
...
+
+  - Made the column `createdAt` on table `Phrase` required. This step will fail if there are existing NULL values in that column.
+
+*/
+
+UPDATE "Phrase"
+  SET "createdAt" = COALESCE(data.timestamp, '2024-02-01')
+FROM (
+  SELECT ph.id, e.timestamp FROM "Phrase" AS ph
+  LEFT JOIN "GlossEvent" AS e ON e."phraseId" = ph.id
```
arrocke, line 12: This will set created at dates on all phrases before making it required. If there is no gloss event associated with the phrase (which will be true for all english glosses), then we fall back to february 1, which is one month before the first gloss event in the database

tycebrown:
@arrocke I'm having trouble migrating the database. I ran `pg_restore -f test.sql seed.dump` and, based on the generated sql file, it appears that the seed file contains phrase entries with null  `createdAt` fields. Since these phrases are inserted when the db is reseeded (_after_ the prisma migrations are run), wouldn't `createdAt` field already be required?

If I understand this correctly, I believe the database needs to be reseeded with phrases which already have the `createdAt` date set.

arrocke:
woops, thanks for catching that. It should be good now

tycebrown (approved):
Looks good! :+1:

Pertempto:
I'll try to review this evening

Pertempto:
> all queries filter out deleted queries

deleted glosses?

Pertempto:
Ah, deleted phrases.

Pertempto (approved):
It seems to work well 👍🏽

PR #407, opened 2024-07-05
feat: create and remove multi word glosses

### LINKED ISSUE #377: Phrase glosses
We need to support aglutinave languages that need to combine words to create a gloss. These languages can express in one word what in other languages require multiple words.

# stories
- #376 
- #384 
- #361

arrocke:
## What has changed

* Routes to add and delete phrases
* UI to select words to link in a phrase, link and unlink words, show words linked to current word
* Keyboard shortcuts to select words, link and unlink

## Connected Issues

closes #377 

## QA Steps

1. Use the checkbox and Ctrl+Enter the select/deselect words
2. Use the link words button and Ctrl+L to link selected words
3. Use the unlink words button and Ctrl+U to unlink selected words
4. Confirm that linked words can't be selected to form new phrases
5. Confirm that linked words show background when one of the words in the phrase is highlighted

## Post-Deployment

- [x] Nothing required

### DIFF
--- a/packages/web/src/features/translation/TranslateWord.tsx
+++ b/packages/web/src/features/translation/TranslateWord.tsx
@@ -323,14 +323,14 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
                     );
                   }}
                   onKeyDown={(e) => {
-                    if (e.metaKey || e.ctrlKey) return;
+                    if (e.metaKey || e.altKey) return;
                     switch (e.key) {
                       case 'Enter': {
                         e.preventDefault();
                         if (e.shiftKey) {
                           const prev = root.current?.previousElementSibling;
                           prev?.querySelector('input')?.focus();
-                        } else if (e.altKey) {
+                        } else if (e.ctrlKey) {
                           if (!isMultiWord) {
                             onSelect?.();
                           }
@@ -344,6 +344,8 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
                         break;
                       }
                       case 'Escape': {
+                        if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey)
+                          return;
                         onChange({ approved: false });
                         break;
                       }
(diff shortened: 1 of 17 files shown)


packages/web/src/features/translation/TranslateWord.tsx
```diff
@@ -246,13 +323,17 @@ const TranslateWord = forwardRef<TranslateWordRef, TranslateWordProps>(
...
                     switch (e.key) {
                       case 'Enter': {
                         e.preventDefault();
                         if (e.shiftKey) {
                           const prev = root.current?.previousElementSibling;
                           prev?.querySelector('input')?.focus();
+                        } else if (e.ctrlKey) {
+                          if (!isMultiWord) {
+                            onSelect?.();
+                          }
```
tycebrown, line 336: Just a minor point: Ctrl + Enter will only select the focused word if the word's gloss input is also focused. Would it make more sense to move the `onKeyDown` for Ctrl + Enter onto the parent `li` tag?

tycebrown (approved):
Here are my questions:
1. Do we want the user to be able to link non-adjacent words?
2. At the current moment, the gloss of a phrase is not preserved when that phrase is unlinked. Do we want to change that behavior?
3. Do you think we should open an issue for some sort of help page, with all the ctrl shortcuts listed?
4. A thought: Do you think it would be a better user experience if, instead of having to use a checkbox to select words, they could just use shift+click? It seems that would make the ui and controls a little simpler.

arrocke:
Here are my thoughts:
1. Yes. This will especially be true for greek where words can be reordered in interesting ways. This system is designed to handle that by allowing you to select any set of words in the verse to link together
2. If you entered a gloss for words that you linked together, I don't think that's going to be useful if you change your mind that they aren't associated. We will definitely revisit this if we get feedback from translators
3. Yep, this is a good idea
4. I actually started with shift+click, the main issue with that is discoverability. By showing a checkbox on hover and focus, you make it clear to the user that selection is possible. It's the same problem with using Enter to approve glosses. Users don't know that's possible unless you show them.

PR #417, opened 2024-08-10
feat: show progress in the translation view

### LINKED ISSUE #411: show progress in the translation view
## Motivation
Being able to see the progress for the current book of the Bible solves two problems:
* Translators are motivated to make progress. It serves as a little reward to make the progress bar move
* Translators can quickly see if there are any words they've missed.

## Scope
* In the translation toolbar, there is a teal progress bar that spans the full width of the screen. It tracks the completion percentage by number of words approved in the current book.
* On hover, the bar expands down (without moving the content below the toolbar) and reveals a label that specifies the number of completed words out of the total with a percent
* The placement of the words changes based on whether there is sufficient room in the teal section of the bar for the entire text.

Image

## Plan

- [x] Figma design (@arrocke)
- [ ] Implement

tycebrown:
## What has changed
- The `TranslationView `displays a book progress bar directly below the `TranslationToolbar`

## Connected Issues

closes #411 

## QA Steps
1. Go to a verse and hover over the progress bar. Confirm that the approved words to total words count is accurate. 
2. Approve a gloss for a word and hover over the progress bar again. Confirm that the approved words count has gone up by one and the total words count has remained the same. 
3. Link two words and give them an approved gloss. Confirm that the approved words count has gone up by two and the total words count has remained the same.
4. In VSCode navigate to `TranslationToolbar`. Manually replace the value of the local variable `percentageFull` variable to 25%.  Reload the app.
5. Confirm that the progress text is now displayed within the teal section of the progress bar.
## Post-Deployment

- [ ] Nothing required


packages/web/src/features/translation/TranslationProgressBar.tsx
```diff
@@ -0,0 +1,69 @@
...
+    12; // 12px for the right/end margin;
+
+  const [textFitsInside, setTextFitsInside] = useState(true);
+
+  useEffect(() => {
+    const { current: progressElement } = progressElementRef;
+    if (!progressElement) return;
+    const resizeObserver = new ResizeObserver(() => {
+      setTextFitsInside(textElementWidth <= progressElement.offsetWidth);
+    });
```
arrocke, line 52: We should create the resize observer with useRef, and then observe in the effect and unobserve when the effect is torn down. This prevents the observer from being recreated whenever the effect runs

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -209,10 +216,15 @@ export default function TranslationView() {
...
-    onSettled: (_, __, ___, context) => {
+    onSettled: (_, __, variables, context) => {
       queryClient.invalidateQueries({
         queryKey: ['verse-glosses', language, verseId],
       });
+      if (variables.state) {
+        queryClient.invalidateQueries({
+          queryKey: ['language-progress', language],
+        });
+      }
```
arrocke: We'll want to add this to the approve all mutation as well

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -70,6 +71,10 @@ function useTranslationQueries(language: string, verseId: string) {
     { enabled: !!selectedLanguage }
   );
 
+  const bookProgressQuery = useQuery(['language-progress', language], () =>
+    apiClient.languages.findProgress(language)
+  );
```
arrocke: We'll want to add a custom api route to get the progress for a specific book. This query is going to be very slow because it has to load the entire Bible. We can also move this into the progress bar itself since we aren't using it elsewhere

packages/web/src/features/translation/TranslationView.tsx
```diff
@@ -461,6 +473,16 @@ export default function TranslationView() {
           }
         }}
       />
+      <TranslationProgressBar
```
arrocke: We should conditionally render this based on whether the user is a member of the language, so that normal readers don't see it

arrocke (changes requested):

packages/web/src/features/translation/TranslationView.tsx

packages/web/src/features/translation/TranslationView.tsx

packages/web/src/features/translation/TranslationView.tsx

packages/web/src/features/translation/TranslationProgressBar.tsx
tycebrown: Done.

tycebrown: I ended up also adding up using `useCallback` to capture the value of `textElementWidth`

packages/api-client/src/books.ts
```diff
@@ -0,0 +1,15 @@
+import { GetBookProgressResponseBody } from '@translation/api-types';
+import ApiClient from './client';
+
+export default class Books {
+  constructor(private readonly client: ApiClient) {}
+
+  findProgress(
```
arrocke: We can make this a method in the languages module and name it `findBookProgress`

packages/web/src/features/translation/TranslationProgressBar.tsx
```diff
@@ -27,17 +42,27 @@ export default function TranslationProgressBar(
...
     12; // 12px for the right/end margin;
 
+  const onResize = useCallback(
+    (progressElement: HTMLDivElement) => {
+      setTextFitsInside(textElementWidth <= progressElement.offsetWidth);
+    },
+    [textElementWidth]
+  );
+  const resizeObserverRef = useRef(
+    new ResizeObserver(([progressElementObserver]) =>
```
arrocke: I'm sorry, I gave you bad advice here. I forgot that the resize observer callback has to change when the textElementWidth changes. The proper way to design that is what you had.

arrocke (changes requested):
I pushed a commit that improves the performance of the progress query by about 50%. I also noticed this when testing:

Screen Shot 2024-08-13 at 10 46 06 PM

arrocke (approved):

